PR #379, opened 2026-06-28
fix(planner): guard insertFixItTask against duplicate task IDs

### LINKED ISSUE #358: Guard insertFixItTask against duplicate task IDs
## Finding
`insertFixItTask` can overwrite an existing task ID.

## Evidence
- `packages/franken-planner/src/core/dag.ts:77-90` rejects duplicate IDs in `addTask`.
- `packages/franken-planner/src/core/dag.ts:115-129` does not perform the same duplicate check before `newNodes.set(fixTask.id, fixTask)` and `newEdges.set(fixTask.id, ...)`.

## Impact
Recovery insertion can silently replace an existing task and its dependencies, corrupting the plan graph instead of failing fast.

## Suggested fix
Add the same duplicate-ID guard used by `addTask`, throwing `DuplicateTaskError` when `fixTask.id` already exists. Add a regression test.

## Review source
Found during principal-level code review of origin/main @ 1f819ef9f239137df6977bfbe57442d256a1d2a6.

djm204:
Closes #358

## Problem
`PlanGraph.insertFixItTask` inserted `fixTask` into the node/edge maps without checking whether its id already existed — unlike `addTask`, which throws `DuplicateTaskError`. A fix-it task reusing an existing id silently overwrote that task and its dependencies, corrupting the plan graph instead of failing fast.

## Fix
Added the same duplicate-ID guard used by `addTask` to `insertFixItTask`, throwing `DuplicateTaskError` when `fixTask.id` already exists.

## Tests
Added a regression test asserting `insertFixItTask` throws `DuplicateTaskError` for a colliding id. Per-package build, test (200 passing), and typecheck all pass.

### DIFF
--- a/packages/franken-planner/src/core/dag.ts
+++ b/packages/franken-planner/src/core/dag.ts
@@ -116,6 +116,9 @@ export class PlanGraph {
     if (!this._nodes.has(failedTaskId)) {
       throw new TaskNotFoundError(failedTaskId);
     }
+    if (this._nodes.has(fixTask.id)) {
+      throw new DuplicateTaskError(fixTask.id);
+    }
     const failedDeps = this._edges.get(failedTaskId) ?? new Set<TaskId>();
 
     const newNodes = new Map(this._nodes);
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #383, opened 2026-06-28
fix(franken-observer): handle PostMortemGenerator file write errors gracefully

### LINKED ISSUE #72: PostMortemGenerator crashes on file write errors (disk full, read-only)
## Severity: MEDIUM
**Area:** \`franken-observer\` — PostMortemGenerator

## Summary

\`generate()\` calls \`mkdir()\` and \`writeFile()\` without try-catch. If the output directory is read-only or disk is full, the unhandled rejection crashes the process.

## Evidence

- \`franken-observer/src/incident/PostMortemGenerator.ts:72-80\`

## Acceptance Criteria

- [ ] Wrap file operations in try-catch
- [ ] Log and degrade gracefully on write failure

djm204:
Closes #72

## Problem
`PostMortemGenerator.generate()` called `mkdir()` and `writeFile()` without a try-catch (`PostMortemGenerator.ts:72-80`). A read-only output directory or a full disk produced an unhandled promise rejection that crashed the very agent loop the post-mortem was meant to diagnose.

## Fix
- Wrapped directory creation and file write in a try-catch.
- On failure, the error is logged via `console.warn` (matching the package's `[Tag]` convention) and `generate()` returns `null` — degrading gracefully rather than propagating the crash.
- Return type is now `Promise<string | null>`. No external caller consumes the value as a non-null string.

## Tests (TDD)
Added two tests that trigger real filesystem failures (no fragile ESM mocking):
- mkdir failure via a file-as-parent (`ENOTDIR`)
- writeFile failure via a directory pre-created at the target path (`EISDIR`)

Both assert `generate()` does not throw, returns `null`, and logs a warning.

## Verification (per-package, in worktree)
- `npm run build` — pass
- `npm test` — 446 passed (33 files)
- `npm run typecheck` — pass

### DIFF
--- a/packages/franken-observer/src/incident/PostMortemGenerator.test.ts
+++ b/packages/franken-observer/src/incident/PostMortemGenerator.test.ts
@@ -1,7 +1,7 @@
-import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { join } from 'node:path'
 import { tmpdir } from 'node:os'
-import { existsSync, readFileSync, rmSync } from 'node:fs'
+import { existsSync, readFileSync, rmSync, writeFileSync, mkdirSync } from 'node:fs'
 import { randomUUID } from 'node:crypto'
 import { TraceContext } from '../core/TraceContext.js'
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #439, opened 2026-06-29
fix(cli): honor non-interactive init

### LINKED ISSUE #398: fix(cli): honor init --non-interactive without prompting
## Finding
`frankenbeast init --non-interactive` is parsed and advertised, but the init command still enters prompting/interactive paths.

## Evidence
- `--non-interactive` is parsed into `initNonInteractive` in `packages/franken-orchestrator/src/cli/args.ts`.
- `handleInitCommand` does not use that flag.
- Init can still prompt for encrypted backend passphrase and runs `runInteractiveInit` unless verify/repair.

## Impact
Headless setup/CI can hang despite explicitly passing `--non-interactive`.

## Suggested fix
Make non-interactive init fail fast with actionable missing-config errors, or accept all required values through flags/env and never call interactive prompt paths when the flag is set.

djm204:
## Summary
- make `frankenbeast init --non-interactive` verify existing init state/config and fail fast instead of prompting
- add regression coverage to ensure non-interactive init does not call prompt paths when required config is missing

Closes #398

## Test plan
- `npm exec --workspace packages/franken-orchestrator -- vitest run tests/unit/cli/init-command.test.ts`
- `npm exec --workspace packages/franken-orchestrator -- vitest run tests/unit/init/init-engine.test.ts tests/unit/init/init-verify.test.ts tests/unit/cli/init-command.test.ts`
- `npm run --workspace packages/franken-orchestrator typecheck` (fails on pre-existing `makeTokenSpend` export errors in observer adapters)
- `npm run --workspace packages/franken-orchestrator test` (fails on the same pre-existing observer adapter `makeTokenSpend` runtime failures)

### DIFF
--- a/packages/franken-orchestrator/src/cli/init-command.ts
+++ b/packages/franken-orchestrator/src/cli/init-command.ts
@@ -33,6 +33,24 @@ export async function handleInitCommand(options: InitCommandOptions): Promise<vo
     return;
   }
 
+  if (options.args.initNonInteractive) {
+    const verification = await verifyInit({
+      configFile: options.paths.configFile,
+      stateStore,
+    });
+    if (!verification.ok) {
+      throw new Error(
+        [
+          'Cannot run init non-interactively because required init configuration is missing or incomplete:',
+          ...verification.messages.map((message) => `- ${message}`),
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

Issue #475, opened 2026-07-01
Governor DANGEROUS_PATTERNS heuristic: verb-aware matching + split-flag detection


djm204:
#### File and Lines
* **File**: [`packages/franken-mcp-suite/src/adapters/governor-adapter.ts` lines 61-73](file:///home/pfkagent/dev/frankenbeast/packages/franken-mcp-suite/src/adapters/governor-adapter.ts#L61-L73)

#### Current Code Behavior
The governor adapter uses simple substring regexes to catch dangerous actions:
```typescript
const DANGEROUS_PATTERNS = [
  /delete/i,
  /drop/i,
  /truncate/i,
  /destroy/i,
  /remove.*all/i,
  /force.*push/i,
  /reset.*hard/i,
  /rm\s+-rf/i,
  /format/i,
  /wipe/i,
  /purge/i,
];
```
This matches benign words (such as `src/dropdown.tsx` or `formatMessage()`), causing false-positive blockages. Conversely, dangerous commands with separate options (such as `rm -r -f` or `rm --recursive --force`) bypass the `rm\s+-rf` rule.

#### Proposed Solution
1. **Word boundary anchors**: Anchor verbs where possible, e.g. `\bdelete\b`, `\btruncate\b`.
2. **Split flag detection**: Expand the `rm` regex to match recursive and force flags in any order:
   ```typescript
   /rm\s+(?:-\w*[rf]\w*\s+|--force\s+|--recursive\s+)*-\w*[rf]\w*/i
   ```

---

### Test Plan
1. Run governor adapter tests:
   ```bash
   npx vitest run packages/franken-mcp-suite/src/adapters/governor-adapter.test.ts
   ```

PR #581, opened 2026-07-05
fix(governor): make trigger evaluators actually fire and 'denied' reachable

### LINKED ISSUE #490: fix(governor): BudgetTrigger/SkillTrigger never fire — adapter passes RationaleBlock lacking their context fields
## Gap
`GovernorCritiqueAdapter.evaluateTriggers` passes the raw `RationaleBlock` to every registered trigger evaluator, but `BudgetTrigger` reads `context.tripped` and `SkillTrigger` reads `context.requiresHitl`/`context.isDestructive` — none of which exist on `RationaleBlock`. Both triggers therefore always return `{ triggered: false }`: the documented gating (governor README Quick Start's "Blocked" branch, README.md:44-66) is unreachable and the gate always approves.

## Evidence
- `packages/franken-governor/src/gateway/governor-critique-adapter.ts:69-75` — `evaluator.evaluate(rationale)`.
- `packages/franken-governor/src/triggers/skill-trigger.ts:13-16` — requires `SkillTriggerContext { skillId, requiresHitl, isDestructive }`.
- `packages/franken-governor/src/triggers/budget-trigger.ts:14` — requires `context.tripped`.
- `packages/franken-types/src/rationale.ts:7-13` — `RationaleBlock` has none of these fields.

The generic-erasure of `TriggerEvaluator<T>` in the registry hides the type mismatch.

## Proposed fix
Either (a) have the adapter construct the correct per-trigger contexts (budget state from an injected budget source, skill metadata from the rationale/skill registry), or (b) redefine the trigger contract to evaluate `RationaleBlock` and derive their signals from it. Update the governor README Quick Start to match whichever contract wins, and add a test that a tripped budget / destructive skill actually produces a non-approved verdict through `verifyRationale`.

### LINKED ISSUE #491: fix(mcp-suite): governor 'denied' decision is unreachable — mapping requires severity=critical but no trigger emits it
## Gap
The mcp-suite governor adapter can never return `denied`. `mapSeverityToDecision` returns `denied` only for `severity === 'critical'`, but the only registered trigger (`SkillTrigger` via the dangerous-pattern path) always emits `severity: 'high'`. docs/walkthrough-mcp-suite.md ("What the full-cycle test actually checks", step 4) claims `rm -rf` produces a `denied` row in `governor_log` with exit 1 — the actual value is `review_recommended`.

## Evidence
- `packages/franken-mcp-suite/src/adapters/governor-adapter.ts:42` — `if (severity === 'critical') return 'denied';`
- `packages/franken-governor/src/triggers/skill-trigger.ts` — always `severity: 'high'`.
- `packages/franken-mcp-suite/tests/full-cycle.integration.test.ts:109` — asserts `['denied','review_recommended']`, i.e. the test itself hedges.

## Proposed fix
Make `denied` reachable: emit `critical` for destructive/dangerous-pattern matches (or map `high` from the dangerous-pattern trigger to `denied`). Tighten the full-cycle test to assert the specific decision, and align the walkthrough text. Coordinate with the tool-argument plumbing issue and #475.

djm204:
## Summary
Two governance defects from the 2026-07-04 docs-vs-reality audit, both verified live at HEAD before fixing.

### #490 — triggers never fired
`GovernorCritiqueAdapter.evaluateTriggers` passed the raw `RationaleBlock` to every evaluator, but `SkillTrigger` reads `requiresHitl`/`isDestructive` and `BudgetTrigger` reads `tripped` — none exist on `RationaleBlock`, so both always returned `triggered: false` and the gate always approved. The generic erasure of `TriggerEvaluator<T>` hid the mismatch.

**Fix:** the adapter now constructs per-trigger contexts:
- `SkillTrigger` → `{skillId, requiresHitl, isDestructive}` derived from `rationale.selectedTool` via a new injectable `SkillMetadataSource`.
- `BudgetTrigger` → circuit-breaker state from a new injectable `BudgetStateSource`.
- A trigger whose context cannot be constructed is **skipped explicitly** (documented) instead of being fed a mistyped object; custom evaluators still receive the `RationaleBlock`.
- `createGovernor` accepts the two optional sources; additive API only — the mcp-suite call path that constructs `SkillTriggerContext` directly is untouched.
- Governor README Quick Start now wires the sources, so its `Blocked` branch is actually reachable (it previously could never fire).

### #491 — 'denied' unreachable
`mapSeverityToDecision` (mcp-suite governor adapter) returns `denied` only for severity `critical`, but `SkillTrigger` always emitted `high` — so `rm -rf` produced `review_recommended`, contradicting the walkthrough's documented `denied`.

**Fix:** `SkillTrigger` emits `critical` for destructive skills (`high` remains for HITL-only), making destructive/dangerous-pattern matches map to a hard `denied` across the shared governor path (hooks, public check tool, central ADR-038 gate). The full-cycle integration test assertion is tightened from `['denied','review_recommended']` to exactly `'denied'`; a payload-only dangerous command (`run_shell` + `rm -rf …`) is also asserted `denied`. Word-order pattern variants (e.g. `push --force`) remain #475's scope.

## Testing
- New `GovernorCritiqueAdapter` suite: destructive skill → rejected, HITL skill → rejected, benign → approved, tripped budget → rejected, untripped → approved, explicit-skip cases (no source / no selectedTool / unknown skill).
- `npx turbo run test typecheck --filter=@franken/governor --filter=@fbeast/mcp-suite --force` — 12/12 tasks green.

Closes #490
Closes #491

### DIFF
--- a/packages/franken-governor/README.md
+++ b/packages/franken-governor/README.md
@@ -40,11 +40,19 @@ import * as readline from 'node:readline/promises';
 const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
 const readlineAdapter = { question: (prompt: string) => rl.question(prompt) };
 
-// Wire up the governor with triggers
+// Wire up the governor with triggers. SkillTrigger and BudgetTrigger need
+// context sources: skillMetadata supplies HITL/destructive flags for the
+// rationale's selectedTool, and budgetState supplies the circuit-breaker
+// state. A trigger whose source is missing is skipped (it cannot fire).
 const governor = createGovernor({
   readline: readlineAdapter,
   memoryPort: { recordDecision: async (trace) => console.log('Audit:', trace) },
   evaluators: [new BudgetTrigger(), new SkillTrigger()],
+  skillMetadata: {
+    getSkillMetadata: (skillId) =>
+      skillId === 'deploy-prod' ? { requiresHitl: true, isDestructive: true } : undefined,
+  },
+  budgetState: { getBudgetState: () => ({ tripped: false, spendUsd: 0, limitUsd: 50 }) },
   projectId: 'my-project',
   operatorName: 'alice',
 });
@@ -58,6 +66,8 @@ const result = await governor.verifyRationale({
   timestamp: new Date(),
 });
 
+// 'deploy-prod' is flagged destructive above, so the SkillTrigger fires and
+// the operator is prompted; a rejection (or non-TTY default) lands here:
 if (result.verdict === 'approved') {
   console.log('Proceeding with execution');
 } else {
--- a/packages/franken-governor/src/gateway/governor-critique-adapter.ts
+++ b/packages/franken-governor/src/gateway/governor-critique-adapter.ts
@@ -5,7 +5,28 @@ import type { ApprovalChannel } from './approval-channel.js';
 import { ApprovalGateway, type AuditRecorder } from './approval-gateway.js';
 import type { SignatureVerifier } from '../security/signature-verifier.js';
 import type { TriggerEvaluator } from '../triggers/trigger-evaluator.js';
+import { BudgetTrigger, type BudgetTriggerContext } from '../triggers/budget-trigger.js';
+import { SkillTrigger, type SkillTriggerContext } from '../triggers/skill-trigger.js';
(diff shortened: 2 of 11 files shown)


djm204:
@codex review

Issue #640, opened 2026-07-05
fix(orchestrator): stop swallowing exceptions silently in pr-creator


djm204:
### Detailed Triage & Code Quality Inspection

The `PrCreator` class wraps several asynchronous calls (LLM prompts and git executions) in catch-all `try-catch` blocks that return `null` on error, completely discarding context.

#### 1. generateCommitMessage
* **File**: [`packages/franken-orchestrator/src/closure/pr-creator.ts` lines 128-130](file:///home/pfkagent/dev/frankenbeast/packages/franken-orchestrator/src/closure/pr-creator.ts#L128-L130)
* **Snippet**:
  ```typescript
  } catch {
    return null;
  }
  ```

#### 2. generatePrDescription
* **File**: [`packages/franken-orchestrator/src/closure/pr-creator.ts` lines 178-180](file:///home/pfkagent/dev/frankenbeast/packages/franken-orchestrator/src/closure/pr-creator.ts#L178-L180)
* **Snippet**:
  ```typescript
  } catch {
    return null;
  }
  ```

#### 3. tryGeneratePrFromLlm
* **File**: [`packages/franken-orchestrator/src/closure/pr-creator.ts` lines 369-371](file:///home/pfkagent/dev/frankenbeast/packages/franken-orchestrator/src/closure/pr-creator.ts#L369-L371)
* **Snippet**:
  ```typescript
  } catch {
    return null;
  }
  ```

#### Impact
This silences API authentication failures, rate limits, network timeouts, or git parsing errors. The execution flow continues without a PR being created, but developers have no logs to diagnose why the PR was skipped.

#### Proposed Solution
Ensure the caught `error` is recorded using the logger before returning the fallback value:
```typescript
} catch (error) {
  logger?.warn('PrCreator: generateCommitMessage failed', {
    error: error instanceof Error ? error.message : String(error)
  });
  return null;
}
```

---

### Test Plan
1. Run typecheck to verify compiling:
   ```bash
   npx turbo run typecheck --filter=franken-orchestrator
   ```
2. Run `PrCreator` tests:
   ```bash
   npx vitest run packages/franken-orchestrator/tests/unit/closure/pr-creator.test.ts
   ```

PR #842, opened 2026-07-06
fix(critique): declare acorn parser deps to unbreak main (CI red)

djm204:
## Urgent: main is currently red
A clean checkout of `main` does not build. CI's last runs (#818 release, #837) fail with:
```
src/evaluators/logic-loop.ts(2,26): error TS2307: Cannot find module 'acorn-typescript'
@franken/critique#build: ERROR ... exited (2)
```

## Root cause
`packages/franken-critique/src/evaluators/logic-loop.ts` imports both `acorn` and `acorn-typescript`, but the package declared **neither**, and `acorn-typescript` was absent from the lockfile entirely. It only ever worked on machines where a stale `node_modules` happened to carry it. This also breaks the **release/publish pipeline** — every publishable package's `prepublishOnly` runs the build, so releases fail too.

## Fix
Declare both parser deps in `franken-critique` dependencies:
- `acorn: ^8.17.0` (directly imported on lines 1 & 3 — declaring it avoids relying on transitive hoisting, the same phantom-dep fragility)
- `acorn-typescript: ^1.4.13`

Lockfile change is scoped to adding `acorn-typescript` (acorn was already present transitively).

## Verification
- Clean `npm ci` + `npx turbo run build` → **10/10 green** (was failing on `@franken/critique#build`).
- `@franken/critique` test + lint → green.

## Note
An in-flight PR (#840, nominally about issue #640) also carries an `acorn-typescript` declaration bundled in. This dedicated PR unblocks main immediately without waiting on that unrelated change; #840 will need only a trivial rebase afterward (its acorn hunk becomes a no-op). It also adds `acorn` explicitly, which #840 does not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3290,8 +3290,9 @@
       }
     },
     "node_modules/acorn": {
-      "version": "8.16.0",
-      "dev": true,
+      "version": "8.17.0",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.17.0.tgz",
+      "integrity": "sha512-xRQbDb9BnwDafYNn6Vwl839DYVjqXYb1XVGtWAZ1kcDc6iwAL4hg3B1dZlRiuENFeO2H53gFG3in621AdERVAg==",
       "license": "MIT",
       "bin": {
         "acorn": "bin/acorn"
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

djm204:
Superseded by #840, which merged the acorn-typescript declaration and unbroke main (verified: a clean critique build now passes on origin/main @ e49fa8dc). 

Residual (optional, minor): logic-loop.ts also imports `acorn` directly, which #840 does not declare — it resolves transitively via acorn-typescript's own dependency, so it builds, but declaring the directly-imported package explicitly would harden against a future hoisting change. Not worth a standalone PR; noting for the record. Closing this as redundant.

PR #854, opened 2026-07-06
fix(orchestrator): make sharp optional so the published CLI runs standalone

djm204:
## Publish-integrity blocker (found during Phase-2 boundary proof)
Installed from a packed tarball (as a real `npm install -g` would), the flagship `frankenbeast` CLI **crashes on every invocation** before it can print anything:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'sharp' imported from .../@franken/orchestrator/dist/logging/beast-logger.js
```

## Root cause
`beast-logger.ts` imported `sharp` at the top level, but **no package declared it** — only the repo *root* does (`sharp ^0.34.5`). In the monorepo it resolves via hoisting; published/installed standalone it is absent, and the top-level import runs at module load, so the whole CLI dies.

`sharp` is used only to render a **cosmetic PNG banner**, and the code already has an ASCII fallback (`buildFallbackBanner`, gated by `shouldRenderGraphicBanner` / `FRANKENBEAST_PLAIN_BANNER`).

## Fix
- Load `sharp` lazily via `await import('sharp')` **inside the existing `try`**, so its absence is caught and degrades to the ASCII banner instead of crashing.
- Declare `sharp` in `optionalDependencies` (`^0.34.5`) so it installs when the platform supports it, but a missing/failed native install never breaks `npm install` or the CLI.

Optional (not top-level) is the right call: sharp is a large native module and forcing it as a hard dependency would bloat the CLI and break installs on unsupported platforms just for a logo.

## Verification (offline packed-artifact install, outside the monorepo)
- **Before:** `frankenbeast --help` → `ERR_MODULE_NOT_FOUND`, exit 1.
- **After (sharp omitted via `--omit=optional`):** `frankenbeast --help` → full help, **exit 0**.
- `beast-logger` + `run` unit tests green (112).

Part of proving the publish pipeline yields a runnable CLI. Pairs with #844 (files allowlist); a CI pack-and-install smoke test to catch this class automatically is planned as a follow-up guard.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -614,7 +614,7 @@
     },
     "node_modules/@img/colour": {
       "version": "1.1.0",
-      "dev": true,
+      "devOptional": true,
       "license": "MIT",
       "engines": {
         "node": ">=18"
@@ -6549,7 +6549,7 @@
     },
     "node_modules/sharp": {
       "version": "0.34.5",
-      "dev": true,
+      "devOptional": true,
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #858, opened 2026-07-06
fix(governor): guard signature buffer lengths

### LINKED ISSUE #849: fix(governor): prevent timingSafeEqual length mismatch crash in SignatureVerifier
### Description
`SignatureVerifier.verify` checks length constraints on the hex string format before buffer decoding:
```typescript
if (signature.length !== 64)
```
If a 64-character signature contains non-hex characters, buffer decoding fails or returns mismatched lengths. When passed to Node's `timingSafeEqual`, it throws an unhandled `RangeError / TypeError: Input buffers must have the same length`, causing server/request crashes.

### Proposed Solution
Validate hex format and decode buffers to compare actual buffer byte lengths before invoking `timingSafeEqual()`.

---

### Test Plan
1. Run signature validation tests:
   ```bash
   npx turbo run test --filter=franken-governor
   ```

djm204:
## Summary
- Reject non-hex approval signatures before decoding buffers
- Compare decoded HMAC buffer lengths before timingSafeEqual
- Add a regression test for 64-character non-hex signatures returning false instead of throwing

## Test Plan
- npm --prefix packages/franken-governor test -- tests/unit/security/signature-verifier.test.ts
- npx turbo run test --filter=@franken/governor
- npx turbo run typecheck --filter=@franken/governor
- npx turbo run build --filter=@franken/governor

Closes #849

### DIFF
--- a/packages/franken-governor/src/security/signature-verifier.ts
+++ b/packages/franken-governor/src/security/signature-verifier.ts
@@ -26,11 +26,12 @@ export class SignatureVerifier {
 
   verify(payload: string, signature: string): boolean {
     const expected = this.sign(payload);
-    if (expected.length !== signature.length) return false;
+    if (!/^[0-9a-f]{64}$/i.test(signature)) return false;
 
-    return timingSafeEqual(
-      Buffer.from(expected, 'hex'),
-      Buffer.from(signature, 'hex'),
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1275, opened 2026-07-08
fix(critique): detect ghost dependencies in re-exports

### LINKED ISSUE #1165: fix(critique): GhostDependencyEvaluator misses `export ... from` re-exports, passing code that ships ghost dependencies
**Finding**
The ghost-dependency scanner only recognizes `import` and `require` keywords, so a module whose only reference to a missing package is a re-export (`export { x } from 'pkg'` / `export * from 'pkg'` — the standard barrel-file pattern) sails through with `verdict: 'pass', score: 1` from an evaluator whose findings are marked `severity: 'critical'`.

**Evidence**
- `packages/franken-critique/src/evaluators/ghost-dependency.ts:104-119` — the scan loop only triggers on `startsWithKeyword(content, i, 'import')` and `startsWithKeyword(content, i, 'require')`; there is no `export` handler, and the `'pkg'` specifier after `from` is consumed by the generic string-skip branch at lines 99-102 (`skipStringLiteral`) without being recorded.
- `ghost-dependency.ts:147-151` — dynamic `import('pkg')` is also skipped, but at least that skip is explicit and commented:
  ```ts
  // Ignore import.meta, dynamic import(...), and non-declaration object keys ...
  if (content[i] === '.' || content[i] === '(' || content[i] === ':') {
    return null;
  }
  ```
  The `export ... from` gap has no comment and no test — `tests/unit/evaluators/ghost-dependency.test.ts` contains zero `export` cases.

**Impact**
Generated code that re-exports from an undeclared package passes the critique gate, and the ghost dependency surfaces later as a runtime module-resolution crash in the consuming project — precisely the failure class this critical-severity evaluator exists to block.

**Suggested fix**
Treat `export` followed by a `from '<spec>'` clause the same as static `import` (reuse `readStaticImportSpecifier`, which already parses the `from` clause), add test cases for `export { x } from`, `export * from`, and make an explicit detect-or-documented-skip decision for dynamic `import('pkg')`.

djm204:
## Summary
- scan static re-export declarations for package specifiers alongside imports
- add regression coverage for named and namespace re-exports
- document the current dynamic import skip with an explicit regression test

## Test Plan
- npm test -- --run tests/unit/evaluators/ghost-dependency.test.ts -t "re-exports|dynamic import" (fails before fix, passes after)
- npm test -- --run tests/unit/evaluators/ghost-dependency.test.ts
- npm run lint --workspace @franken/critique
- npm run build --workspace @franken/critique

Closes #1165

### DIFF
--- a/packages/franken-critique/src/evaluators/ghost-dependency.ts
+++ b/packages/franken-critique/src/evaluators/ghost-dependency.ts
@@ -110,6 +110,15 @@ function extractDependencySpecifiers(content: string): string[] {
       continue;
     }
 
+    if (startsWithKeyword(content, i, 'export')) {
+      const reexported = readStaticImportSpecifier(content, i + 'export'.length);
+      if (reexported) {
+        specifiers.push(reexported.value);
+        i = reexported.endIndex;
+      }
+      continue;
+    }
+
     if (startsWithKeyword(content, i, 'require')) {
       const required = readRequireSpecifier(content, i + 'require'.length);
       if (required) {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1276, opened 2026-07-08
fix(orchestrator): surface beast daemon dispatch failures

### LINKED ISSUE #1164: fix(orchestrator): BeastDaemonDispatchAdapter swallows daemon failures as "not a beast request"
**Finding**
`BeastDaemonDispatchAdapter.handle` catches every catalog-fetch failure (daemon down, wrong `baseUrl`, 401 bad operator token) and returns `null` — the exact signal used for "the user didn't ask to launch a beast" — so daemon outages are indistinguishable from a non-matching chat message.

**Evidence**
- `packages/franken-orchestrator/src/chat/beast-daemon-dispatch-adapter.ts:62-67`:
  ```ts
  let definitions: BeastDefinitionSummary[];
  try {
    definitions = await this.listDefinitions();
  } catch {
    return null;
  }
  ```
- `beast-daemon-dispatch-adapter.ts:205-212` — `request()` throws on any non-2xx (`Beast daemon request failed: ${status}`), so auth and connectivity errors all land in that catch.
- `null` is also the "no launch intent" return at lines 58-60 and 69-71; the adapter is wired into the live chat server at `packages/franken-orchestrator/src/http/chat-server.ts:288`.
- Inconsistent policy in the same class: the equivalent failure one call later (`resultFromProgress` → `listDefinitions()` at line ~99) propagates uncaught.

**Impact**
A user typing "spawn a martin beast" against a misconfigured or unreachable daemon gets an ordinary LLM chat reply with no indication anything failed. A rotated operator token or dead daemon is undetectable from the chat surface — it just looks like the intent matcher never triggered. Silent dead end for the operator, and a fail-open shape (error → treated as benign no-match) in the dispatch path.

**Suggested fix**
Distinguish "no match" from "daemon unreachable": catch, log the underlying error, and return a dispatch result telling the user the beast daemon is unavailable (or rethrow and let the chat engine render an error turn), matching the propagate-on-failure behavior of `resultFromProgress`.

djm204:
## Summary
- propagate Beast daemon catalog failures for launch-like chat requests instead of converting them to a no-match null result
- keep ordinary chat turns and reachable-daemon no-match launch requests on the existing null path
- add targeted BeastDaemonDispatchAdapter coverage for failure vs no-match behavior

## Test Plan
- npm --prefix packages/franken-orchestrator test -- tests/unit/chat/beast-daemon-dispatch-adapter.test.ts
- npm run build
- npm --prefix packages/franken-orchestrator run typecheck
- npm --prefix packages/franken-orchestrator run lint

Closes #1164

### DIFF
--- a/packages/franken-orchestrator/src/chat/beast-daemon-dispatch-adapter.ts
+++ b/packages/franken-orchestrator/src/chat/beast-daemon-dispatch-adapter.ts
@@ -59,12 +59,7 @@ export class BeastDaemonDispatchAdapter {
       return null;
     }
 
-    let definitions: BeastDefinitionSummary[];
-    try {
-      definitions = await this.listDefinitions();
-    } catch {
-      return null;
-    }
+    const definitions = await this.listDefinitions();
     const definition = this.matchDefinition(input, definitions);
     if (!definition) {
       return null;
--- a/packages/franken-orchestrator/tests/unit/chat/beast-daemon-dispatch-adapter.test.ts
+++ b/packages/franken-orchestrator/tests/unit/chat/beast-daemon-dispatch-adapter.test.ts
@@ -106,4 +106,38 @@ describe('BeastDaemonDispatchAdapter', () => {
     })).resolves.toBeNull();
     expect(fetchMock).not.toHaveBeenCalled();
   });
+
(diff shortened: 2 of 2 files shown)


djm204:
@codex review

PR #1306, opened 2026-07-09
fix(planner): fail parallel planner on dangling dependencies

### LINKED ISSUE #937: fix(planner): fail ParallelPlanner on dangling dependencies instead of reporting success
## Description
`ParallelPlanner` breaks out of its execution loop when no tasks are ready, then returns `status: 'completed'` regardless of whether every task ran. This is known for cycles, but the same behavior occurs for dangling dependencies: if task `T` depends on missing task `D`, `T` never becomes ready, the loop exits once independent tasks finish, and the planner reports success with partial results.

## Affected locations
- `packages/franken-planner/src/planners/parallel.ts` lines 20-28
- `packages/franken-planner/src/planners/parallel.ts` line 88

## Impact
Plans with missing dependency references can silently skip tasks while reporting successful completion.

## Recommendation
Validate dependency references before execution or assert `completedIds.size === tasks.length` before returning success. Return/throw a structured failed result when dangling dependencies prevent progress.

djm204:
## Summary
- fail `ParallelPlanner` with a structured failed result when a task depends on a missing task
- add a regression that proves dangling dependencies do not execute partial plans or report success

## Test Plan
- npm --prefix packages/franken-planner test -- tests/unit/planners/parallel.test.ts
- npm --prefix packages/franken-planner run typecheck
- npm --prefix packages/franken-planner run build
- npm --prefix packages/franken-planner run lint
- npm --prefix packages/franken-planner test

Closes #937

### DIFF
--- a/packages/franken-planner/src/planners/parallel.ts
+++ b/packages/franken-planner/src/planners/parallel.ts
@@ -60,6 +60,21 @@ export class ParallelPlanner implements PlanningStrategy {
       throw new RecursionDepthExceededError(depth);
     }
 
+    // Validate dangling dependency references before sorting/executing so a
+    // malformed plan cannot run a partial independent subset and be reported as
+    // success. Cycles remain a structural graph error from topoSort below.
+    const danglingDependency = findDanglingDependency(graph);
+    if (danglingDependency) {
+      return {
+        status: 'failed',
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1323, opened 2026-07-09
fix(observer): report ADR rule exceptions by rule

### LINKED ISSUE #1223: fix(observer): report ArchitecturalAdherence rule exceptions by rule
## Description
`ArchitecturalAdherenceEval` executes every ADR rule with `rules.filter(r => !r.check(output))` and does not catch per-rule exceptions. A single buggy or dependency-backed `ADRRule.check` can therefore abort the whole eval through `EvalRunner`, returning only a generic `Eval threw an unexpected error` message and losing the specific rule name/description that operators need to fix the architecture gate.

## Affected locations
- `packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.ts:3-8` defines `ADRRule` callbacks without an error contract.
- `packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.ts:24-32` calls each rule's `check(output)` inside `Array.filter` with no per-rule isolation.
- `packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.ts:38-48` only reports names for rules that return `false`; thrown rules never appear in `violatedRules`.
- `packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.test.ts:46-74` covers returning `false`, but has no regression for a throwing ADR rule.

## Evidence / reasoning path
1. Register an ADR rule such as `{ name: 'no-layer-violation', description: '...', check: () => { throw new Error('parser failed') } }`.
2. Run `ArchitecturalAdherenceEval` through `EvalRunner`.
3. The exception escapes line 31 before `violatedRules` is built.
4. `EvalRunner` catches it as a generic eval failure, so the result does not identify which ADR rule failed or allow other rules to continue.

This weakens reliability of architecture gates: one fragile rule can mask all other violations and produce poor diagnostics instead of a structured failed rule.

## Recommendation
Evaluate rules one at a time inside a try/catch. Treat a thrown rule as a failed rule with details such as `{ rule: name, error: message }`, continue evaluating remaining rules, and include the rule name in `reason`/`details`.

## Acceptance criteria
- A throwing ADR rule produces a structured `status: 'fail'` result instead of relying on the runner's generic exception wrapper.
- `details.violatedRules` includes the throwing rule name.
- `details` also preserves the thrown message for diagnostics.
- Other ADR rules are still evaluated when one rule throws.
- Regression tests cover one throwing rule plus one normal failing/passing rule.

Duplicate check: checked local issue cache and targeted live searches for `ArchitecturalAdherenceEval`, `ADRRule`, `rule.check`, `deterministic eval`, and `eval exception`; adjacent issues #1210 and #1184 cover critique/governor evaluator exception isolation, not this observer deterministic eval's per-rule diagnostics.

djm204:
## Summary
- isolate ArchitecturalAdherence ADR rule checks so one thrown rule does not abort the eval
- report thrown rules as violated rules with per-rule diagnostic messages
- add regression coverage for one throwing rule plus normal failing and passing rules

## Test Plan
- npm test -- ArchitecturalAdherence.test.ts
- npm test -- --filter=@franken/observer
- npm run build
- npm run typecheck

Closes #1223

### DIFF
--- a/packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.test.ts
+++ b/packages/franken-observer/src/evals/deterministic/ArchitecturalAdherence.test.ts
@@ -72,6 +72,42 @@ describe('ArchitecturalAdherenceEval', () => {
       })
       expect(result.reason).toContain('TypeScript output must not use the `any` type')
     })
+
+    it('reports throwing rules by rule and continues evaluating remaining rules', async () => {
+      const evaluatedRules: string[] = []
+      const throwingRule: ADRRule = {
+        name: 'parse-adr-links',
+        description: 'ADR references must be parseable',
+        check: () => {
+          evaluatedRules.push('parse-adr-links')
+          throw new Error('ADR parser failed')
+        },
+      }
+      const passingRule: ADRRule = {
+        name: 'passing-rule',
+        description: 'A passing rule should still run after an earlier exception',
+        check: () => {
+          evaluatedRules.push('passing-rule')
+          return true
+        },
+      }
+
+      const result = await runner.run(ev, {
+        output: `const x: any = 1`,
+        rules: [throwingRule, noAnyRule, passingRule],
+      })
+
+      expect(result.status).toBe('fail')
+      expect(result.reason).toContain('[parse-adr-links] ADR references must be parseable')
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1367, opened 2026-07-09
fix(mcp): reject unknown critique evaluators

### LINKED ISSUE #1255: fix(mcp): reject unknown critique evaluator names
## Description
`fbeast_critique_evaluate` accepts a comma-separated `evaluators` argument, but unknown evaluator names are silently ignored all the way through the critique adapter. If every requested evaluator is unknown, the adapter falls back to the default evaluator set and returns a successful-looking critique result instead of telling the caller that their evaluator selection was invalid.

That makes typos such as `logicloop` (instead of `logic-loop`) or stale evaluator names hard to detect: the tool appears to honor the request, but actually runs a different set of checks.

## Affected locations
- `packages/franken-mcp-suite/src/servers/critique.ts:22-31` documents `evaluators` as a comma-separated selector and forwards the parsed strings from `splitCsvArg(...)` without validating them.
- `packages/franken-mcp-suite/src/adapters/critique-adapter.ts:88-97` maps requested names through `evaluatorMap`, filters unknown names out, then falls back to the full default evaluator set when the filtered selection is empty.
- `packages/franken-mcp-suite/src/servers/critique.test.ts:17-48` covers valid evaluator forwarding, but there is no negative coverage for unknown-only or partially unknown evaluator selections.

1. Call `fbeast_critique_evaluate` with `content: "..."` and `evaluators: "logicloop"`.
2. `splitCsvArg` returns `["logicloop"]` and the server passes it to `critique.evaluate`.
3. `resolveEvaluators(["logicloop"])` finds no matching evaluator, filters to `[]`, and falls back to `LogicLoopEvaluator`, `ComplexityEvaluator`, and `ConcisenessEvaluator`.
4. The response reports a normal verdict/score, giving no indication that the caller's requested evaluator name was ignored.

## Impact
Low-risk reliability/DX bug: callers can believe a narrow evaluator selection ran when the system actually ran defaults, which makes automated critique results misleading and can hide broken prompt/tool wiring.

## Recommendation
Validate evaluator names at the MCP boundary or adapter boundary. Unknown evaluator names should return a clear tool error naming the invalid values. For partially invalid lists, either reject the whole request or report the invalid names explicitly before running only known evaluators.

## Acceptance criteria
- Passing only unknown evaluator names to `fbeast_critique_evaluate` returns an error instead of silently running defaults.
- Passing a mix of known and unknown evaluator names is handled deterministically and documented (preferably rejected with the unknown names listed).
- Regression tests cover unknown-only and partially unknown evaluator lists.
- Valid evaluator selection (`logic-loop,complexity`) keeps working.

Duplicate check: read the local open-issues cache and targeted live searches for `critique criteria evaluators empty CSV`, `splitCsvArg critique`, `fbeast_critique_evaluate`, `unknown evaluator`, `logic-loop`, and `evaluator validation`. No open issue covered invalid critique evaluator names; adjacent critique issues in the cache cover parser/complexity heuristics, not MCP evaluator selection validation.

djm204:
## Summary
- reject unknown fbeast_critique_evaluate evaluator names instead of falling back to defaults
- report all invalid evaluator names while documenting the supported values
- add regression coverage for unknown-only and partially unknown evaluator lists

## Test Plan
- npm --workspace @franken/mcp-suite test -- src/servers/critique.test.ts
- npm --workspace @franken/mcp-suite run typecheck
- npm run build
- npm run lint:any

Closes #1255

### DIFF
--- a/packages/franken-mcp-suite/src/adapters/critique-adapter.ts
+++ b/packages/franken-mcp-suite/src/adapters/critique-adapter.ts
@@ -86,15 +86,17 @@ export function createCritiqueAdapter(): CritiqueAdapter {
   };
 
   function resolveEvaluators(names?: string[]) {
-    const selected = (names && names.length > 0 ? names : defaultEvaluators())
-      .map((name) => evaluatorMap.get(name))
-      .filter((evaluator): evaluator is Evaluator => evaluator !== undefined);
+    const requestedNames = names && names.length > 0 ? names : defaultEvaluators();
+    const unknownNames = requestedNames.filter((name) => !evaluatorMap.has(name));
 
-    return selected.length > 0 ? selected : [
-      new LogicLoopEvaluator(),
-      new ComplexityEvaluator(),
-      new ConcisenessEvaluator(),
-    ];
+    if (unknownNames.length > 0) {
+      const noun = unknownNames.length === 1 ? 'evaluator' : 'evaluators';
+      throw new Error(
+        `Unknown critique ${noun}: ${unknownNames.join(', ')}. Expected one of: ${defaultEvaluators().join(', ')}`,
+      );
+    }
+
+    return requestedNames.map((name) => evaluatorMap.get(name)!);
   }
 }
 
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #1386, opened 2026-07-09
fix(orchestrator): reject invalid analytics time windows

### LINKED ISSUE #1217: fix(orchestrator): reject invalid Analytics timeWindow filters
## Description
The Analytics API accepts arbitrary `timeWindow` query values and silently treats invalid values as unbounded/all-time filters. The web UI only offers `24h`, `7d`, `30d`, and `all`, but the endpoint is callable directly and the frontend forwards filter values as query strings without validation. A typo or stale deep link such as `?timeWindow=30days` returns all events instead of rejecting the invalid filter or preserving the intended bounded window.

## Affected locations
- `packages/franken-orchestrator/src/http/routes/analytics-routes.ts:55-62` — `readFilters()` forwards any non-empty `timeWindow` string into the analytics service.
- `packages/franken-orchestrator/src/analytics/sqlite-analytics-service.ts:519-526` — `cutoffFor()` returns `null` for unrecognized values, and `matchesFilters()` interprets `null` as no cutoff/all time.
- `packages/franken-web/src/lib/analytics-api.ts:95-102` — the web client serializes the filter value directly, so copied URLs or future UI state can hit the silent fallback path.

1. Call an analytics endpoint with an invalid time window, for example `/api/analytics/events?timeWindow=30days` or `/api/analytics/summary?timeWindow=yesterday`.
2. `readFilters()` accepts the value because it only checks truthiness.
3. `cutoffFor()` fails to match `/^(\d+)([hd])$/` and returns `null`.
4. `matchesFilters()` applies no timestamp cutoff, so the response is indistinguishable from `timeWindow=all`.

## Impact
Operators can believe they are looking at a bounded recent Analytics view while actually seeing all historical audit/governor/cost/beast events. That can skew failure counts, cost totals, and incident triage when an invalid query parameter comes from a hand-edited URL, bookmark, stale client state, or future UI regression.

## Recommendation
Validate `timeWindow` at the route boundary and reject unsupported values with a 400 structured error, or normalize to a documented allowlist (`24h`, `7d`, `30d`, `all`) before calling the service. Add route/service regression tests that invalid values do not silently expand to all time.

## Acceptance criteria
- Invalid Analytics `timeWindow` values return a clear 400 response or another explicit validation error instead of all-time results.
- Supported values (`24h`, `7d`, `30d`, `all`) continue to work.
- Tests cover at least one invalid value such as `30days` and prove it is not treated as `all`.

Duplicate check: searched live open issues for `analytics timeWindow`, `time window`, `Last 24h`, `readFilters`, and `cutoffFor`; only related Analytics issues found were session picker stability and pagination, not invalid time-window validation.

djm204:
## Summary
- Validate Analytics `timeWindow` filters at the HTTP route boundary.
- Return a structured 400 `invalid_time_window` error for unsupported values.
- Add unit coverage proving invalid values do not call the analytics service or silently expand to all-time results.

## Test Plan
- [x] `cd packages/franken-orchestrator && npm test -- tests/unit/http/analytics-routes.test.ts`
- [x] `npm run typecheck --workspace @franken/orchestrator`
- [x] `npm run build --workspace @franken/orchestrator`
- [x] `npm run lint --workspace @franken/orchestrator` (passes with pre-existing warnings)
- [x] `git diff --check`

Closes #1217

### DIFF
--- a/packages/franken-orchestrator/src/http/routes/analytics-routes.ts
+++ b/packages/franken-orchestrator/src/http/routes/analytics-routes.ts
@@ -1,4 +1,5 @@
 import { Hono } from 'hono';
+import type { Context } from 'hono';
 import type { AnalyticsFilters, AnalyticsOutcome, AnalyticsPageRequest, AnalyticsService } from '../../analytics/types.js';
 
 export interface AnalyticsRouteDeps {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1401, opened 2026-07-09
fix(observer): validate LoopDetector numeric options

### LINKED ISSUE #1232: fix(observer): validate LoopDetector numeric options
## Description
`LoopDetector` accepts numeric tuning options directly and never validates that they are finite, positive, and internally meaningful. Invalid values can either disable detection silently or produce nonsensical detections.

## Affected locations
- `packages/franken-observer/src/incident/LoopDetector.ts:35-46` — constructor stores `windowSize`, `repeatThreshold`, `similarityThreshold`, `maxGapBetweenRepetitions`, and `historyLimit` without bounds checks.
- `packages/franken-observer/src/incident/LoopDetector.ts:67-90` — detection math assumes positive window/repetition sizes (`minLength`, `latestPatternStart`, `cursor`) but is driven by unchecked options.
- `packages/franken-observer/src/incident/LoopDetector.ts:102-110` — gap search uses unchecked `maxGapBetweenRepetitions`, so negative/NaN values change the scan window instead of being rejected.
- `packages/franken-observer/src/incident/LoopDetector.test.ts:49-71` and `:128-142` — custom-option coverage only exercises valid positive values; there are no invalid-option regressions.

1. Construct `new LoopDetector({ windowSize: 0, repeatThreshold: 3 })`.
2. Call `check('anything')`; `minLength` is `0`, `pattern` is `[]`, and `windowMatches()` vacuously succeeds, so the detector can report an empty detected pattern unrelated to span history.
3. Construct `new LoopDetector({ repeatThreshold: 0 })`; the same `minLength = 0` path lets the first check report a detected loop with `repetitions: 1` even though zero repetitions should be invalid configuration.
4. Construct `new LoopDetector({ similarityThreshold: Number.NaN })`; fuzzy matching can never satisfy `>= NaN`, silently disabling near-match loop detection.
5. Construct `new LoopDetector({ windowSize: -1 })`; `latestPatternStart` and `slice()` math operate on negative windows rather than failing fast, producing unpredictable no-op behavior.

## Impact
Bad configuration turns a reliability guard into a source of false positives or false negatives. Operators can either receive empty-pattern loop detections that interrupt healthy runs, or lose loop detection entirely without any explicit configuration error. Because these options are public API surface, consumers need deterministic validation failures instead of silent degradation.

## Recommendation
Validate constructor options before storing them:
- `windowSize`, `repeatThreshold`, and `historyLimit` should be finite positive integers.
- `maxGapBetweenRepetitions` should be a finite non-negative integer.
- `similarityThreshold` should be finite and between `0` and `1` inclusive.
- Preserve the existing `historyLimit = max(configured/default, spanNeededForDetection)` behavior after validating the configured value.
Add unit tests for zero, negative, non-integer, and `NaN` values plus one regression that `windowSize: 0` cannot emit an empty-pattern detection.

## Acceptance criteria
- Invalid `LoopDetectorOptions` throw a clear `RangeError`/`TypeError` during construction.
- `LoopDetector.check()` cannot emit detections with an empty `detectedPattern` caused by invalid window/repetition sizes.
- Tests cover invalid `windowSize`, `repeatThreshold`, `similarityThreshold`, `maxGapBetweenRepetitions`, and `historyLimit` values.
- Existing valid custom-option tests still pass.

Duplicate check: read the local open-issue cache and recent PM fallback claims; targeted live searches for `LoopDetector windowSize repeatThreshold historyLimit`, `LoopDetector invalid options`, `similarityThreshold`, and `maxGapBetweenRepetitions` returned no open duplicate. Reviewed adjacent #1020, which covers handler exception isolation only, not constructor option bounds.

djm204:
## Summary
- validate LoopDetector numeric options at construction time
- reject invalid window/repetition/history/gap/threshold values with clear RangeError messages
- add regression coverage for zero, negative, fractional, NaN, and infinite option values

## Test Plan
- [x] `npm --workspace=@franken/observer test -- LoopDetector.test.ts`
- [x] `npm --workspace=@franken/observer test`
- [x] `npm --workspace=@franken/types run build && npm --workspace=@franken/observer run typecheck && npm --workspace=@franken/observer run build`

Closes #1232

### DIFF
--- a/packages/franken-observer/src/incident/LoopDetector.test.ts
+++ b/packages/franken-observer/src/incident/LoopDetector.test.ts
@@ -70,6 +70,53 @@ describe('LoopDetector', () => {
     })
   })
 
+  describe('option validation', () => {
+    it.each([
+      ['windowSize', { windowSize: 0 }],
+      ['windowSize', { windowSize: -1 }],
+      ['windowSize', { windowSize: 1.5 }],
+      ['windowSize', { windowSize: Number.NaN }],
+      ['repeatThreshold', { repeatThreshold: 0 }],
+      ['repeatThreshold', { repeatThreshold: -1 }],
+      ['repeatThreshold', { repeatThreshold: 1.5 }],
+      ['repeatThreshold', { repeatThreshold: Number.POSITIVE_INFINITY }],
+      ['historyLimit', { historyLimit: 0 }],
+      ['historyLimit', { historyLimit: -1 }],
+      ['historyLimit', { historyLimit: 1.5 }],
+      ['historyLimit', { historyLimit: Number.NaN }],
+    ])('rejects invalid positive integer option %s=%j', (optionName, options) => {
+      expect(() => new LoopDetector(options)).toThrow(new RangeError(`${optionName} must be a finite positive integer`))
+    })
+
+    it.each([
+      ['maxGapBetweenRepetitions', { maxGapBetweenRepetitions: -1 }],
+      ['maxGapBetweenRepetitions', { maxGapBetweenRepetitions: 1.5 }],
+      ['maxGapBetweenRepetitions', { maxGapBetweenRepetitions: Number.NaN }],
+    ])('rejects invalid non-negative integer option %s=%j', (optionName, options) => {
+      expect(() => new LoopDetector(options)).toThrow(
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1443, opened 2026-07-10
fix(observer): tolerate invalid post-mortem timestamps

### LINKED ISSUE #1146: fix(observer): tolerate invalid post-mortem interrupt timestamps
## Description
`PostMortemGenerator.generate()` promises that post-mortem reporting failures degrade gracefully and return `null`, but it builds the markdown content before entering the protected write block. `generateContent()` calls `new Date(signal.timestamp).toISOString()` directly, so a malformed, non-finite, or out-of-range interrupt timestamp throws `RangeError: Invalid time value` before the `try/catch` that handles report persistence failures. That means a bad interrupt signal can crash the loop-reporting path instead of producing a fallback report or returning `null`.

## Affected locations
- `packages/franken-observer/src/incident/PostMortemGenerator.ts:25-27` derives `detectedAt` with `new Date(signal.timestamp).toISOString()` without validating the timestamp.
- `packages/franken-observer/src/incident/PostMortemGenerator.ts:81-93` calls `generateContent()` before the `try` block, so timestamp conversion failures bypass the existing graceful error handling.
- `packages/franken-observer/src/incident/PostMortemGenerator.test.ts:94-100` only covers a valid timestamp; the disk/write failure tests do not cover content-generation failures.

1. Construct an `InterruptSignal` with the required shape but an invalid numeric timestamp, e.g. `{ timestamp: Number.NaN }` or an out-of-range number.
2. Call `await new PostMortemGenerator({ outputDir }).generate(trace, signal)`.
3. `generate()` calls `generateContent()` before entering the `try` block.
4. `generateContent()` executes `new Date(signal.timestamp).toISOString()` and throws `RangeError: Invalid time value`.
5. The promise rejects instead of returning `null`/writing a best-effort report, contradicting the documented "reporting failure never crashes the loop" behavior.

## Impact
This is low-risk but reliability-visible: malformed/corrupt loop detector or interrupt payloads can take down the diagnostic path exactly when the agent is already trying to stop an infinite loop. Operators lose the post-mortem and may see an unhandled reporter exception instead of a graceful fallback.

## Recommendation
Validate `signal.timestamp` before formatting it. Either fall back to `new Date().toISOString()`/`String(signal.timestamp)` for invalid values, or move content generation inside the `try` and return `null` on content-generation failures too. Add a regression test with `timestamp: Number.NaN` or an out-of-range timestamp asserting `generate()` does not throw and records the degraded behavior.

## Acceptance criteria
- `PostMortemGenerator.generate()` does not reject when `signal.timestamp` is invalid or out of range.
- Generated markdown either uses a safe fallback timestamp or the method returns `null` with a warning.
- A regression test covers invalid timestamp input and proves the documented graceful-degradation contract.

Duplicate check: searched the local open-issue cache for `PostMortemGenerator`, `post-mortem`, `invalid timestamp`, `detectedAt`, and `InterruptSignal`; targeted live `gh issue list --search` returned no open duplicate. Existing #571 covers filename sanitization/path traversal only, not invalid timestamp handling.

djm204:
## Summary
- Validate interrupt timestamps by checking the constructed Date before ISO formatting
- Fall back to the existing invalid timestamp marker for non-finite and out-of-range values
- Add regression coverage for out-of-range post-mortem timestamps

## Test Plan
- npm --workspace=@franken/observer test -- src/incident/PostMortemGenerator.test.ts
- npm --workspace=@franken/types run build
- npm --workspace=@franken/observer run typecheck
- npm --workspace=@franken/observer run build
- npm run lint:any
- git diff --check

Closes #1146

### DIFF
--- a/packages/franken-observer/src/incident/PostMortemGenerator.test.ts
+++ b/packages/franken-observer/src/incident/PostMortemGenerator.test.ts
@@ -120,6 +120,14 @@ describe('PostMortemGenerator', () => {
       const content = gen.generateContent(trace, signal)
       expect(content).toContain('2023-11-14') // ISO date for that Unix ms
     })
+
+    it('uses a safe fallback for out-of-range timestamps', () => {
+      const gen = new PostMortemGenerator({ outputDir })
+      const trace = makeTrace()
+      const content = gen.generateContent(trace, makeSignal(trace.id, { timestamp: Number.MAX_VALUE }))
+
+      expect(content).toContain('**Detected at:** Invalid timestamp')
+    })
   })
 
   describe('generate()', () => {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1472, opened 2026-07-10
test(ci): parse workflow YAML assertions

### LINKED ISSUE #1230: test(ci): actually parse workflow YAML in CI workflow tests
## Description
`tests/unit/ci-workflow.test.ts` has a test named "is valid YAML (parseable by node)", but it never parses `.github/workflows/ci.yml`. The assertion only checks that the file is non-empty and contains `name:`, so malformed GitHub Actions YAML can still pass the repository's CI workflow regression test.

## Affected locations
- `tests/unit/ci-workflow.test.ts:22-27` — the "valid YAML" test comments that it should not throw, but only checks `content.length` and `content.includes('name:')`.
- `tests/unit/ci-workflow.test.ts:33-55` — follow-up checks use broad regex/string matching, so they can still pass against syntactically invalid workflow text.
- `.github/workflows/ci.yml:1-22` — this is the workflow under test, but no YAML parser validates it before the string assertions run.

1. Read `tests/unit/ci-workflow.test.ts:22-27`: the test title promises parseability, but there is no YAML parser/import and no parse call.
2. The test would still pass if the workflow file retained `name:` while containing malformed indentation or invalid structure elsewhere.
3. The repository already has release workflow parsing tracked separately in #1065, but the CI workflow test has the same weak-assertion pattern and is not covered by that release-specific ticket.

## Impact
Maintainers can break the GitHub Actions workflow syntax while the local regression suite still reports the CI workflow test as green. That delays detection until GitHub tries to load the workflow, and it weakens confidence in future changes to CI triggers, setup-node options, or command wiring.

## Recommendation
Parse `.github/workflows/ci.yml` with a real YAML parser in the test before the string/structure assertions. Prefer asserting the parsed object has the expected `name`, `on.push.branches`, `on.pull_request.branches`, setup-node step, npm install step, and Turbo command rather than relying only on substring matches.

## Acceptance criteria
- The CI workflow test fails on syntactically invalid YAML even if the file still contains `name:`.
- The parsed workflow object is used for at least the top-level trigger/job assertions.
- The existing checks for Node 22, npm cache, `npm ci`, checkout, and `npx turbo run build test lint` remain covered.
- The release-please workflow parsing work tracked by #1065 remains separate and is not conflated with this CI workflow test fix.

Duplicate check: read `/home/pfkagent/.hermes/state/frankenbeast_open_issues_cache.json` and searched live GitHub for `ci-workflow valid YAML parseable`, `workflow YAML parse`, and `CI workflow test weak assertion`. Adjacent #1065 covers parsing `release-please.yml`; #1222 covers CI not running the root Vitest suite. Neither covers `tests/unit/ci-workflow.test.ts` claiming CI YAML parseability without parsing.

djm204:
## Summary
- Adds real YAML parsing for `.github/workflows/ci.yml` in the root CI workflow regression test.
- Converts top-level trigger/job checks to assert against the parsed workflow object instead of only broad string matches.
- Adds an invalid-YAML regression proving the parser path fails even when `name:` is present.

## Test Plan
- [x] `npm run test:root -- tests/unit/ci-workflow.test.ts --pool=threads`
- [x] `TMPDIR="$PWD/.tmp" npm run test:root -- tests/unit/ci-workflow.test.ts`
- [x] `TMPDIR="$PWD/.tmp" npm run test:root`
- [x] `npm run check:package-manager`
- [x] `TMPDIR="$HERMES_KANBAN_WORKSPACE/tmp" npm run lint:security`
- [x] `TMPDIR="$HERMES_KANBAN_WORKSPACE/tmp" npm run build`
- [x] `TMPDIR="$HERMES_KANBAN_WORKSPACE/tmp" npm run typecheck`

Note: `npm test` was also attempted with `TMPDIR="$HERMES_KANBAN_WORKSPACE/tmp"`, but package-level Vitest transforms failed in this constrained workspace with `Unknown system error -122: write` / disk quota errors before executing tests. The root suite and build/typecheck/security checks above pass.

Closes #1230

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,10 +12,12 @@
       ],
       "devDependencies": {
         "@types/better-sqlite3": "^7.6.13",
+        "@types/js-yaml": "^4.0.9",
         "@types/node": "^25.3.0",
         "@vitest/coverage-v8": "^4.1.9",
         "acorn": "^8.17.0",
         "better-sqlite3": "^12.6.2",
+        "js-yaml": "^4.3.0",
         "sharp": "^0.34.5",
         "turbo": "^2.10.3",
         "typescript": "^5.9.3",
@@ -2832,6 +2834,13 @@
     },
     "node_modules/@types/estree": {
       "version": "1.0.8",
+      "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@types/js-yaml": {
+      "version": "4.0.9",
+      "resolved": "https://registry.npmjs.org/@types/js-yaml/-/js-yaml-4.0.9.tgz",
+      "integrity": "sha512-k4MGaQl5TGo/iipqb2UDG2UwjXziSWkh0uysQelTlJpX1qGlpUZYm8PnO4DxG1qBomtJUdYJ6qR6xdIah10JLg==",
       "dev": true,
       "license": "MIT"
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,12 +12,11 @@
       ],
       "devDependencies": {
         "@types/better-sqlite3": "^7.6.13",
-        "@types/js-yaml": "^4.0.9",
         "@types/node": "^25.3.0",
         "@vitest/coverage-v8": "^4.1.9",
         "acorn": "^8.17.0",
         "better-sqlite3": "^12.6.2",
-        "js-yaml": "^4.3.0",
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

djm204:
@codex review

PR #1474, opened 2026-07-10
test(mcp-suite): assert Windows fbeast shim passthrough

### LINKED ISSUE #1228: test(mcp-suite): assert Windows fbeast passthrough resolves npm shims
## Description
`packages/franken-mcp-suite/src/cli/main.test.ts` claims the `fbeast` passthrough should "use shell on Windows so npm .cmd shims are resolved", but the test currently asserts `shell: false`. That matches the implementation in `packages/franken-mcp-suite/src/cli/main.ts`, so the regression test would stay green even though Windows `fbeast <non-mcp command>` passthrough can fail to resolve the `frankenbeast` npm shim.

## Affected locations
- `packages/franken-mcp-suite/src/cli/main.ts:15-19` calls `spawnSync('frankenbeast', process.argv.slice(2), { stdio: 'inherit', shell: false })` for all non-`mcp` passthrough commands.
- `packages/franken-mcp-suite/src/cli/main.test.ts:46-67` has the test title `uses shell on Windows so npm .cmd shims are resolved`, stubs `process.platform` to `win32`, but expects `expect.objectContaining({ shell: false })`.
- `packages/franken-mcp-suite/src/cli/main.test.ts:24-44` separately checks the default passthrough path also expects `shell: process.platform === 'win32'`, so the Windows-specific assertion is inconsistent with the intended behavior.

## Impact
The test suite documents the desired Windows behavior but asserts the opposite. A Windows user running a linked/global `fbeast network up`, `fbeast beasts catalog`, or any other delegated command can hit an `ENOENT` path because Node does not resolve npm `.cmd` shims when spawning bare commands with `shell: false` on Windows. CI/review would not catch the regression because the dedicated test currently blesses the broken value.

1. Read `packages/franken-mcp-suite/src/cli/main.ts:15-19`: passthrough always uses `shell: false`.
2. Read `packages/franken-mcp-suite/src/cli/main.test.ts:46-67`: the Windows-specific test says npm `.cmd` shims require shell support, then asserts `shell: false`.
3. On Windows, npm-installed CLIs are exposed through `.cmd` shims, so `spawnSync('frankenbeast', ..., { shell: false })` is the fragile path the test title says to avoid.

## Recommendation
Update the passthrough implementation to use `shell: process.platform === 'win32'` (or an equivalent Windows-safe shim resolver) and fix the Windows test to assert `shell: true` when `process.platform` is mocked to `win32`. Keep the non-Windows/default passthrough test asserting no shell on POSIX.

## Acceptance criteria
- `fbeast` non-`mcp` passthrough uses a Windows-safe command resolution path for npm `.cmd` shims.
- The Windows-specific regression test fails against `shell: false` and passes only for the Windows-safe behavior.
- POSIX passthrough continues to avoid an unnecessary shell.
- A CLI smoke or unit regression covers at least one delegated non-`mcp` command path.

Duplicate check: searched the local issue cache and targeted live GitHub issues for `Windows shell fbeast`, `npm .cmd`, `spawnSync shell false`, `fbeast passthrough`, and `main.test windows`; no existing open issue covered the Windows passthrough assertion. Adjacent #950 covers live-bench live-test discoverability and #1128 covers live-bench CLI smoke coverage, not `@fbeast/mcp-suite` Windows passthrough behavior.

djm204:
## Summary
- strengthen the Windows `fbeast` passthrough regression to model a lowercase npm `.cmd` shim
- resolve PATHEXT candidates case-insensitively so mocked Windows tests and real npm shim casing stay aligned
- keep POSIX passthrough on `shell: false`

## Test Plan
- [x] `TMPDIR=/home/pfkagent/dev/resolve-wt/issue-1228/.tmp npx vitest run src/cli/main.test.ts --reporter=verbose --no-color -t 'resolves Windows npm .cmd shims'` (failed before implementation, passed after)
- [x] `TMPDIR=/home/pfkagent/dev/resolve-wt/issue-1228/.tmp npx vitest run src/cli/main.test.ts --reporter=dot --no-color`
- [x] `npm run build --workspace @franken/types`
- [x] `npm run build --workspace @franken/brain`
- [x] `npm run build --workspace @franken/critique`
- [x] `npm run build --workspace @franken/governor`
- [x] `npm run build --workspace @franken/observer`
- [x] `npm run build --workspace @franken/planner`
- [x] `npm run build --workspace @franken/orchestrator`
- [x] `npm --prefix packages/franken-mcp-suite run typecheck`
- [x] `npm --prefix packages/franken-mcp-suite run build`
- [x] `npm --prefix packages/franken-mcp-suite test -- --run src/cli/main.test.ts`

Closes #1228

### DIFF
--- a/packages/franken-mcp-suite/src/cli/main.test.ts
+++ b/packages/franken-mcp-suite/src/cli/main.test.ts
@@ -193,10 +193,10 @@ describe('fbeast main CLI', () => {
     mockExit.mockRestore();
   });
 
-  it('launches Windows .cmd shims through cmd.exe without enabling shell mode', async () => {
+  it('resolves Windows npm .cmd shims for non-mcp passthrough commands', async () => {
     const platformSpy = vi.spyOn(process, 'platform', 'get').mockReturnValue('win32');
     const binDir = tmpDir();
-    const shimPath = join(binDir, 'frankenbeast.CMD');
+    const shimPath = join(binDir, 'frankenbeast.cmd');
     writeFileSync(shimPath, '@echo off\r\n');
     vi.stubEnv('PATH', binDir);
     vi.stubEnv('PATHEXT', '.COM;.EXE;.BAT;.CMD');
--- a/packages/franken-mcp-suite/src/cli/main.ts
+++ b/packages/franken-mcp-suite/src/cli/main.ts
@@ -31,10 +31,12 @@ function getEnvPath(env: NodeJS.ProcessEnv): string {
 }
 
 function windowsCommandCandidates(command: string): string[] {
-  const pathext = (process.env.PATHEXT || '.COM;.EXE;.BAT;.CMD')
+  const pathext = [...new Set((process.env.PATHEXT || '.COM;.EXE;.BAT;.CMD')
     .split(';')
-    .map((ext) => ext.trim())
-    .filter(Boolean);
+    .flatMap((ext) => {
+      const trimmed = ext.trim();
+      return trimmed ? [trimmed, trimmed.toLowerCase(), trimmed.toUpperCase()] : [];
+    }))];
   const hasWindowsPathSeparator = command.includes('/') || command.includes('\\');
   const commandHasExt = win32.extname(command) !== '';
 


djm204:
@codex review

PR #1520, opened 2026-07-10
fix(governor): verify Slack webhooks from raw bytes

### LINKED ISSUE #1046: fix(governor): resolve Slack webhook signature verification encoding mismatch
### Description
In `POST /v1/webhook/slack` inside `app.ts`, the raw Slack request body is parsed using `c.req.text()` rather than `c.req.arrayBuffer()`.
Because string conversion can corrupt or alter multi-byte UTF-8 sequences (such as emojis or special symbols) compared to the raw request bytes Slack originally sent, valid webhooks containing these characters fail signature verification.

### Proposed Solution
Read the raw Slack request body as an ArrayBuffer/Buffer and verify the signature using the raw bytes rather than a parsed string.

---

### Test Plan
1. Run Slack webhook signature verification tests with emoji payloads:
   ```bash
   npx turbo run test --filter=franken-governor
   ```

djm204:
## Summary
- verify Slack webhook signatures against the raw request bytes instead of a decoded text body
- keep parsing derived from the same body after signature verification
- add regression coverage for non-round-trippable raw bytes and emoji Slack form payloads

## Test Plan
- npm test --workspace @franken/governor -- --run tests/unit/server/app.test.ts
- npm run typecheck --workspace @franken/governor
- npm run build --workspace @franken/governor
- npm test --workspace @franken/governor
- npx turbo run test --filter=@franken/governor

Closes #1046

### DIFF
--- a/packages/franken-governor/src/server/app.ts
+++ b/packages/franken-governor/src/server/app.ts
@@ -375,9 +375,10 @@ export function createGovernorApp(options: GovernorAppOptions = {}): Hono {
 
   // POST /v1/webhook/slack — Slack interactive message callback
   app.post('/v1/webhook/slack', async (c) => {
-    // Read the raw body once: signature verification must run over the exact
-    // bytes Slack signed, and the parsed payload is derived from the same text.
-    const rawBody = await c.req.text();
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1547, opened 2026-07-10
fix(mcp-suite): validate init mcpServers shape

### LINKED ISSUE #1176: fix mcp init validate mcpServers shape before writing JSON client config
## Description

The Claude/Gemini init path trusts the existing JSON-client settings shape before merging fbeast MCP servers. In `packages/franken-mcp-suite/src/cli/init.ts:85-103`, `settings['mcpServers']` is cast to `Record<string, unknown>` and reused whenever it is non-nullish:

- `settings = JSON.parse(readFileSync(settingsPath, 'utf-8'))` at lines 85-90 accepts any parsed JSON value/shape.
- `const mcpServers = (settings['mcpServers'] as Record<string, unknown>) ?? {}` at line 93 does not check that the value is a plain object.
- lines 96-101 assign fbeast server entries into that value, then line 120 serializes `settings` back to disk.

If an existing `settings.json` has `mcpServers` as an array, assignment to string keys succeeds in memory but `JSON.stringify` serializes the value as an array and drops the fbeast server properties, while init still prints success. If it is a primitive such as a string, the module runs in strict mode and assignment can throw before any useful recovery message. This makes initialization brittle for partially corrupted or hand-edited client config and can leave operators believing fbeast was installed when no MCP entries were written.

## Affected locations

- `packages/franken-mcp-suite/src/cli/init.ts:85-103`
- `packages/franken-mcp-suite/src/cli/init.ts:120-127`
- Related contrast: hook merging already uses `isObjectRecord()` guards at `packages/franken-mcp-suite/src/cli/init.ts:255-260` and `packages/franken-mcp-suite/src/cli/init.ts:271-308`, but the core `mcpServers` merge does not.

## Recommendation

Validate the parsed settings root and `mcpServers` value before merging:

- If the settings root is not a plain object, fail with an actionable error or replace it only after making a backup.
- If `mcpServers` is absent, create `{}`.
- If `mcpServers` exists but is not a plain non-array object, fail with a clear message instead of silently writing a malformed config.
- Add regression tests for absent, valid-object, array, and primitive `mcpServers` values.

## Acceptance criteria

- `runInit()`/JSON-client init never reports success while fbeast MCP entries are dropped because `mcpServers` was an array.
- Invalid `mcpServers` shapes produce a deterministic, human-readable error and leave the original settings file intact or backed up.
- Tests cover malformed `mcpServers` shapes for both Claude/Gemini JSON-client init.

Duplicate check: searched the local open-issue cache and targeted GitHub search for `mcpServers`, `initJsonClient`, `settings.json`, `fbeast init`, and `MCP config`. The only nearby cached issue was #934, which covers cross-project database config isolation/security, not malformed `mcpServers` shape validation or silent dropped JSON-client entries.

djm204:
## Summary
- Validate existing JSON-client mcpServers before merging fbeast entries
- Fail with a clear actionable error for array/primitive mcpServers instead of reporting success with dropped entries
- Add Claude and Gemini regression coverage for malformed mcpServers values

## Test Plan
- npm test --workspace @franken/mcp-suite -- src/cli/init.test.ts
- npm run typecheck --workspace @franken/mcp-suite
- npm run build --workspace @franken/mcp-suite
- npm run lint:any

Closes #1176

### DIFF
--- a/packages/franken-mcp-suite/src/cli/init.test.ts
+++ b/packages/franken-mcp-suite/src/cli/init.test.ts
@@ -115,6 +115,70 @@ describe('fbeast init', () => {
     expect(readFileSync(join(geminiDir, backups[0]!), 'utf-8')).toBe('{ invalid json');
   });
 
+  it('throws with a clear error and preserves Claude .mcp.json when mcpServers is an array', () => {
+    const root = tmpDir();
+    dirs.push(root);
+    const claudeDir = join(root, '.claude');
+    mkdirSync(claudeDir, { recursive: true });
+    const mcpPath = join(root, '.mcp.json');
+    const original = JSON.stringify({ mcpServers: [] });
+    writeFileSync(mcpPath, original);
+
+    expect(() => runInit({ root, claudeDir, hooks: false, client: 'claude' })).toThrow(
+      '.mcp.json mcpServers must be a JSON object',
+    );
+
+    expect(readFileSync(mcpPath, 'utf-8')).toBe(original);
+  });
+
+  it('throws with a clear error and preserves Claude .mcp.json when mcpServers is primitive', () => {
+    const root = tmpDir();
+    dirs.push(root);
+    const claudeDir = join(root, '.claude');
+    mkdirSync(claudeDir, { recursive: true });
+    const mcpPath = join(root, '.mcp.json');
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1559, opened 2026-07-10
fix(web): summarize disabled module overrides in agent rows

### LINKED ISSUE #984: fix web AgentRow module badge counts disabled config as zero modules
## Description
The Beast agent row badge renders `{Object.values(agent.moduleConfig).filter(Boolean).length} modules`. Dispatch only sends `moduleConfig` when at least one module is disabled, and disabled modules are represented as `false`. As a result, an agent with one disabled module can display `0 modules`, which is misleading and makes disabled-module state look like no module configuration exists.

## Affected locations
- `packages/franken-web/src/pages/beast-dispatch-page.tsx:142-165` — dispatch builds a partial `moduleConfig` and only includes it when some values are `false`.
- `packages/franken-web/src/components/beasts/agent-row.tsx:40-43` — the row badge counts truthy values, so a config containing only disabled modules renders `0 modules`.
- `packages/franken-web/src/components/beasts/agent-detail-readonly.tsx:40-45` — detailed view shows entries with enabled/disabled styling, which contrasts with the row summary.

1. Dispatch a Beast with one module toggle turned off.
2. `submitDefinition` sends a `moduleConfig` object because `hasDisabled` is true; the changed module entry is `false`.
3. `AgentRow` sees `agent.moduleConfig` and renders the badge using `.filter(Boolean).length`.
4. Because the only recorded value is `false`, the badge says `0 modules` even though the row is actually indicating module override state.

## Impact
Low-risk UX bug: operators scanning the Beast list get the wrong summary for module overrides, making it harder to notice that a run has modules disabled.

## Recommendation
Render a summary that matches the semantics, such as `1 disabled module`, `N module overrides`, or compute enabled/disabled counts from the full `MODULE_CONFIG_KEYS` set instead of truthy values in a partial config object.

## Acceptance criteria
- A Beast with one disabled module does not display `0 modules`.
- The row summary clearly distinguishes disabled-module overrides from all-enabled/default config.
- Add a component test or Storybook/example fixture covering a partial `moduleConfig` with false values.

Duplicate check: read `/home/pfkagent/.hermes/state/frankenbeast_open_issues_cache.json` first and searched cache/live issues for `moduleConfig`, `module config`, `module count`, `0 modules`, `AgentRow`, and `Module Toggles`; no open issue covered this row summary bug.

djm204:
## Summary
- replace the AgentRow truthy module badge count with enabled/disabled module override summaries
- add a regression test for partial moduleConfig objects containing disabled modules

## Test Plan
- npm test -- tests/components/beasts/agent-row.test.tsx
- npm run build --workspace=@franken/types
- npm run typecheck
- npm run build
- npm test

Closes #984

### DIFF
--- a/packages/franken-web/src/components/beasts/agent-row.tsx
+++ b/packages/franken-web/src/components/beasts/agent-row.tsx
@@ -15,6 +15,26 @@ function formatTime(iso: string): string {
   return new Date(iso).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
 }
 
+function pluralizeModule(count: number): string {
+  return count === 1 ? 'module' : 'modules';
+}
+
+function formatModuleConfigSummary(moduleConfig: NonNullable<TrackedAgentSummary['moduleConfig']>): string {
+  const values = Object.values(moduleConfig);
+  const disabledCount = values.filter((enabled) => enabled === false).length;
+  const enabledCount = values.filter((enabled) => enabled === true).length;
+
+  if (disabledCount > 0 && enabledCount > 0) {
+    return `${disabledCount} disabled, ${enabledCount} enabled ${pluralizeModule(enabledCount)}`;
+  }
+
+  if (disabledCount > 0) {
+    return `${disabledCount} disabled ${pluralizeModule(disabledCount)}`;
+  }
+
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1561, opened 2026-07-10
fix(web): surface Dashboard initial load failures

### LINKED ISSUE #1196: fix(web): surface Dashboard initial load failures
## Description
The Dashboard page can stay on `Loading dashboard...` forever when its initial snapshot request fails. `DashboardPage` only logs `client.fetchSnapshot()` rejections to the console and never clears `loading` or records a visible error. In secured deployments the fallback SSE subscription cannot rescue this state because `DashboardApiClient.subscribeToDashboard()` opens a raw `EventSource` without an Authorization header and the code comment notes that the stream is not usable behind the operator-token gate.

## Affected locations
- `packages/franken-web/src/pages/dashboard-page.tsx:16-28` — initial `fetchSnapshot()` rejection is handled with `.catch(console.error)`, leaving the Zustand `loading` flag untouched.
- `packages/franken-web/src/stores/dashboard-store.ts:17-33` — `loading` starts as `true` and is only set to `false` by `setSnapshot()`; there is no error state or failed-load transition.
- `packages/franken-web/src/lib/dashboard-api.ts:63-72` — the dashboard SSE path is explicitly documented as unauthenticated and not usable behind the operator-token gate, so a failed initial fetch can leave no authenticated recovery path.

1. Open the Dashboard route while the backend is unavailable, returns 401/500, or rejects `/api/dashboard` because the operator token is missing/expired.
2. `DashboardPage` calls `client.fetchSnapshot()`.
3. The promise rejects and the catch handler only calls `console.error`.
4. Because `setSnapshot()` never runs, `useDashboardStore.loading` remains `true`.
5. The user continues to see only `Loading dashboard...` with no in-app error or retry affordance; behind operator-token auth, the unauthenticated EventSource subscription cannot replace the missing snapshot.

## User impact
Operators cannot tell whether the Dashboard is still loading, unauthorized, offline, or broken. This is a low-risk but user-visible runtime bug that hides actionable backend/auth failures behind an indefinite loading state.

## Recommendation
Track a Dashboard load error in the store/page and set `loading` to `false` when the initial snapshot fails. Render a visible error state with a retry action, and keep using successful snapshots/SSE events to clear the error when a later authenticated fetch succeeds. Consider avoiding the raw SSE subscription behind operator-token auth until ticket-based SSE is wired.

## Acceptance criteria
- A failed initial `fetchSnapshot()` renders a visible Dashboard error instead of leaving `Loading dashboard...` indefinitely.
- The error path clears or retries successfully when a later snapshot fetch succeeds.
- Regression tests cover rejected `fetchSnapshot()` and assert `loading` is cleared with an in-app error.
- Existing dashboard mutation rollback work remains separate from initial-load failure handling.

Duplicate check: searched the local open-issue cache and targeted live GitHub for `dashboard`, `fetchSnapshot`, `console.error`, `initial load`, `Loading dashboard`, `DashboardApiClient`, and `dashboard SSE EventSource auth JSON parse`. Reviewed #1096 (failed settings mutations/rollback), #1081 (SSE snapshot JSON parsing), #688/#577 (dashboard SSE auth), and #1017 (Network logs rail); none cover the initial snapshot rejection leaving the Dashboard stuck in the loading state.

djm204:
## Summary
- Adds an explicit regression assertion that rejected Dashboard initial snapshot loads clear the Zustand loading flag.
- Verifies the visible retryable Dashboard error state does not leave operators stuck at `Loading dashboard...`.

## Test Plan
- npm --prefix packages/franken-web test -- dashboard-page.test.tsx
- npm --prefix packages/franken-types run build && npm --prefix packages/franken-web run typecheck
- npm --prefix packages/franken-web run build

Closes #1196

### DIFF
--- a/packages/franken-web/src/pages/dashboard-page.test.tsx
+++ b/packages/franken-web/src/pages/dashboard-page.test.tsx
@@ -56,6 +56,7 @@ describe('DashboardPage', () => {
     render(<DashboardPage client={client} />);
 
     expect((await screen.findByRole('alert')).textContent).toContain('Unable to load dashboard. HTTP 503');
+    expect(useDashboardStore.getState().loading).toBe(false);
     expect(screen.getByRole('button', { name: 'Retry loading dashboard' })).toBeTruthy();
 
     fireEvent.click(screen.getByRole('button', { name: 'Retry loading dashboard' }));


djm204:
@codex review

PR #1601, opened 2026-07-10
fix(observer): validate BatchAdapter numeric options

### LINKED ISSUE #1227: fix(observer): validate BatchAdapter batch and timer options
## Description
`BatchAdapter` accepts numeric batching options without checking that they are finite positive integers. `maxBatchSize` is copied directly in `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:56-59`, then `flush()` only drains when `this.buffer.length >= this.maxBatchSize` at lines 68-73. If config or caller input supplies `NaN` or `Infinity`, that comparison never becomes true, so traces remain buffered until a manual `drain()`/`stop()` happens. `flushIntervalMs` is also only checked with `options.flushIntervalMs && options.flushIntervalMs > 0` at lines 61-63, so `NaN`, `Infinity`, fractions, or other non-finite/unsafe values can silently disable or misconfigure the safety-net timer.

The tests in `packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts` cover happy-path draining, stop(), and timer cancellation, but there is no regression coverage for invalid numeric options. This makes the adapter brittle when values come from JSON/env/config rather than hand-written TypeScript.

## Affected locations
- `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:56-63` — constructor stores `maxBatchSize` and starts the optional interval without finite/integer/range validation.
- `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:68-83` — invalid `maxBatchSize` values prevent automatic draining and leave traces buffered.
- `packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts:12-207` — no invalid-option regression tests.

1. Construct `new BatchAdapter({ adapter, maxBatchSize: Number.NaN })` or with `Infinity`.
2. Call `flush()` repeatedly.
3. `this.buffer.length >= this.maxBatchSize` is always false for `NaN` and never practically true for `Infinity`, so automatic size draining is disabled and traces stay in memory unless a separate `drain()`/`stop()` path runs.
4. Likewise, invalid `flushIntervalMs` values do not produce an explicit configuration error, so the documented safety-net timer can be silently absent or unsafe.

## Impact
Observer traces can be held indefinitely or lost on process exit when a bad runtime config disables both size-based and timer-based draining. Operators get no early error explaining that batching was misconfigured.

## Recommendation
Validate `maxBatchSize` and `flushIntervalMs` in the constructor before storing/using them. Reject non-finite values, `NaN`, `Infinity`, zero/negative values, and likely non-integers with clear `RangeError` messages. Add regression tests proving invalid values throw and valid values preserve current behavior.

## Acceptance criteria
- `BatchAdapter` rejects invalid `maxBatchSize` values such as `NaN`, `Infinity`, `0`, negative numbers, and fractions.
- `BatchAdapter` rejects invalid `flushIntervalMs` values such as `NaN`, `Infinity`, `0`, negative numbers, and fractions when provided.
- Existing size-drain, timer, `stop()`, and read-through behavior remains unchanged for valid options.
- Unit tests cover both invalid option families and at least one valid boundary value.

Duplicate check: read `/home/pfkagent/.hermes/state/frankenbeast_open_issues_cache.json` and searched live GitHub for `BatchAdapter maxBatchSize`, `BatchAdapter invalid maxBatchSize`, `BatchAdapter invalid`, `flushIntervalMs`, and adjacent observer batching terms. Open #1066 covers a batch race/lost-trace concern and #1019 covers retaining traces after failed drains; neither covers numeric option validation.

djm204:
## Summary
- validates BatchAdapter maxBatchSize as a positive safe integer before buffering
- validates provided flushIntervalMs before starting the safety-net timer
- adds regression coverage for invalid numeric options and valid integer boundaries

## Test Plan
- npm run test --workspace @franken/observer -- BatchAdapter.test.ts
- npm run test --workspace @franken/observer
- npm run lint:any

Closes #1227

### DIFF
--- a/packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts
+++ b/packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts
@@ -36,6 +36,62 @@ function deferred<T = void>(): {
   return { promise, resolve, reject }
 }
 
+// ── options ───────────────────────────────────────────────────────────────────
+
+describe('BatchAdapter — options', () => {
+  it.each([
+    Number.NaN,
+    Infinity,
+    0,
+    -1,
+    1.5,
+    Number.MAX_SAFE_INTEGER + 1,
+  ])('rejects invalid maxBatchSize: %s', maxBatchSize => {
+    expect(() => new BatchAdapter({
+      adapter: new InMemoryAdapter(),
+      maxBatchSize,
+    })).toThrow(RangeError)
+  })
+
+  it.each([
+    Number.NaN,
+    Infinity,
+    0,
+    -1,
+    1.5,
+    Number.MAX_SAFE_INTEGER + 1,
+  ])('rejects invalid flushIntervalMs: %s', flushIntervalMs => {
+    const setIntervalFn = vi.fn()
+
+    expect(() => new BatchAdapter({
+      adapter: new InMemoryAdapter(),
+      flushIntervalMs,
+      setInterval: setIntervalFn,
+    })).toThrow(RangeError)
+    expect(setIntervalFn).not.toHaveBeenCalled()
+  })
+
+  it('accepts valid integer boundaries for batch size and timer options', async () => {
+    const inner = new InMemoryAdapter()
+    const setIntervalFn = vi.fn().mockReturnValue(42 as unknown as ReturnType<typeof globalThis.setInterval>)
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1603, opened 2026-07-10
fix(observer): make SpanRedactor regex rules stateless

### LINKED ISSUE #1189: fix(observer): make SpanRedactor RegExp rules stateless
## Description
`SpanRedactor` accepts `RedactionRule.key` as either a string or a `RegExp` (`packages/franken-observer/src/redaction/SpanRedactor.ts:9-15`) and applies regex rules with `rule.key.test(key)` for every metadata key (`packages/franken-observer/src/redaction/SpanRedactor.ts:100-114`). If a caller supplies a global or sticky regex such as `/^(api|auth)_/g`, `RegExp.test()` mutates `lastIndex`; consecutive matching keys then alternate between match and miss. That lets some metadata keys that should be removed or masked pass through to downstream adapters.

The existing test only covers a non-global regex (`packages/franken-observer/src/redaction/SpanRedactor.test.ts:72-85`), so the stateful-regex path is not guarded.

## Affected locations
- `packages/franken-observer/src/redaction/SpanRedactor.ts:9-15` — public `RedactionRule.key` type accepts any `RegExp`, including stateful global/sticky expressions.
- `packages/franken-observer/src/redaction/SpanRedactor.ts:100-114` — matching calls `rule.key.test(key)` directly without resetting `lastIndex` or cloning the regex.
- `packages/franken-observer/src/redaction/SpanRedactor.test.ts:72-85` — coverage proves plain regex rules work but misses `/g` or `/y` behavior.

1. Configure `new SpanRedactor({ rules: [{ key: /^(api|auth)_/g, action: 'remove' }], ... })`. This is allowed by the current `RegExp` type and mirrors the documented prefix matching example except with a common global flag.
2. Flush a span with metadata `{ api_key: 'k', auth_token: 't', model: 'x' }`.
3. `rule.key.test('api_key')` returns true and leaves `lastIndex` non-zero.
4. The next call `rule.key.test('auth_token')` starts from the previous `lastIndex` and can return false, so `auth_token` remains in the redacted trace.

## Impact
Redaction configuration can appear valid and pass current tests while intermittently leaking sensitive metadata fields whenever users provide stateful regex rules. This is especially brittle because the miss depends on metadata key order and previous matches, so the same rule can redact one secret while leaving the next matching secret intact.

## Recommendation
Make regex matching stateless before each key comparison. Options include resetting `rule.key.lastIndex = 0` before and after `test()`, cloning regex rules without `g`/`y`, or rejecting global/sticky regexes at construction time with a clear error. Add regression coverage for a `/g` rule that must redact consecutive matching metadata keys across one span and repeated flushes.

## Acceptance criteria
- Consecutive matching metadata keys are consistently removed/masked when a redaction rule uses `/g` or `/y`.
- Reusing the same `SpanRedactor` across multiple flushes does not carry regex state between spans/traces.
- Tests cover at least one global regex redaction rule and assert all matching keys are redacted.
- Existing non-global regex and exact-string redaction behavior stays unchanged.

Duplicate check: searched local open-issue cache for `SpanRedactor`, `redactor`, `redaction`, `RegExp`, `lastIndex`, `global regex`, `thoughtBlocks`, and `metadata redaction`; targeted `gh issue list` searches for `SpanRedactor redaction RegExp lastIndex global regex`, `redaction regex metadata rule`, and `global RegExp lastIndex` returned no open duplicates. Nearest open issue #1021 covers clone/downstream mutation semantics, not stateful regex matching.

djm204:
## Summary
- match RegExp redaction rules through fresh regex instances so global/sticky lastIndex state cannot leak between metadata keys or flushes
- add regression coverage for consecutive global RegExp removals across repeated flushes and sticky RegExp masking

## Test Plan
- npm test -- src/redaction/SpanRedactor.test.ts (from packages/franken-observer)
- npm run build --workspace @franken/types && npm run typecheck --workspace @franken/observer
- npm test --workspace @franken/observer

Closes #1189

### DIFF
--- a/packages/franken-observer/src/redaction/SpanRedactor.test.ts
+++ b/packages/franken-observer/src/redaction/SpanRedactor.test.ts
@@ -85,6 +85,53 @@ describe('SpanRedactor — metadata rules', () => {
     expect(meta['model']).toBe('x')
   })
 
+  it('matches global RegExp rules statelessly across consecutive keys and flushes', async () => {
+    const inner = new InMemoryAdapter()
+    const keyPattern = /^(api|auth)_/g
+    const redactor = new SpanRedactor({
+      adapter: inner,
+      rules: [{ key: keyPattern, action: 'remove' }],
+    })
+
+    await redactor.flush(
+      makeTrace([makeSpan({ metadata: { api_key: 'k', auth_token: 't', model: 'x' } })]),
+    )
+    await redactor.flush(
+      makeTrace(
+        [makeSpan({ id: 'span-2', metadata: { api_secret: 's', auth_cookie: 'c', safe: true } })],
+        'trace-2',
+      ),
+    )
+
+    const firstMeta = (await inner.queryByTraceId('trace-1'))!.spans[0].metadata
+    expect(firstMeta).not.toHaveProperty('api_key')
+    expect(firstMeta).not.toHaveProperty('auth_token')
+    expect(firstMeta['model']).toBe('x')
+
+    const secondMeta = (await inner.queryByTraceId('trace-2'))!.spans[0].metadata
+    expect(secondMeta).not.toHaveProperty('api_secret')
+    expect(secondMeta).not.toHaveProperty('auth_cookie')
+    expect(secondMeta['safe']).toBe(true)
+    expect(keyPattern.lastIndex).toBe(0)
+  })
+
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

djm204:
@codex review

PR #1607, opened 2026-07-10
fix(observer): isolate circuit breaker handler failures

### LINKED ISSUE #1180: fix(observer): isolate CircuitBreaker handler failures
## Description

`CircuitBreaker` documents itself as a non-blocking budget guard that "Never throws" (`packages/franken-observer/src/cost/CircuitBreaker.ts:13-16`), but `check()` directly invokes every `limit-reached` handler without guarding handler failures (`packages/franken-observer/src/cost/CircuitBreaker.ts:32-40`). If any observer/HITL notification handler throws, `check()` throws too and later handlers are skipped.

This turns a best-effort spend alert into a control-flow failure in the code that is trying to report budget state, and one bad handler can prevent other listeners from receiving the limit-reached event.

## Affected locations

- `packages/franken-observer/src/cost/CircuitBreaker.ts:13-16` — class-level contract says the guard is non-blocking and never throws.
- `packages/franken-observer/src/cost/CircuitBreaker.ts:32-40` — `check()` calls each registered handler directly on the rising edge.
- `packages/franken-observer/src/cost/CircuitBreaker.test.ts:91-100` — current non-blocking tests only cover spend-over-limit itself, not listener exceptions.

1. Register two `limit-reached` handlers on a `CircuitBreaker`.
2. Make the first handler throw, e.g. `breaker.on('limit-reached', () => { throw new Error('webhook failed') })`.
3. Call `breaker.check(limit + 0.01)`.
4. Actual: the thrown handler escapes from `check()`, contradicting the non-blocking contract, and later handlers are not invoked.
5. Expected: `check()` should keep returning a `CircuitBreakerResult`; handler errors should be isolated/logged/aggregated without preventing other handlers from running.

## Recommendation

Wrap handler execution inside the rising-edge loop so a failing listener cannot make `check()` throw or starve subsequent listeners. Add regression tests for:

- a throwing handler does not make `check()` throw;
- later handlers still run after an earlier handler fails;
- existing rising-edge/no-alert-fatigue behavior remains intact.

## Acceptance criteria

- `CircuitBreaker.check()` remains non-throwing when a `limit-reached` handler throws.
- All registered handlers are attempted on the rising edge even if one fails.
- Regression coverage exists in `packages/franken-observer/src/cost/CircuitBreaker.test.ts` for handler exception isolation.
- The behavior is documented if handler errors are logged, swallowed, or exposed through an explicit optional error hook.

## Duplicate check

Checked the local open-issue cache and recent PM fallback claims for `CircuitBreaker`, `limit-reached`, handler, non-blocking, `Never throws`, `spendUsd`, and `limitUsd`. Targeted live searches for `CircuitBreaker handler never throws limit-reached` and `CircuitBreaker limitUsd spendUsd NaN negative` returned no open duplicate. Related #1123 covers a test timing smell around WebhookNotifier/CircuitBreaker assertions, not handler exception isolation.

djm204:
## Summary
- isolate CircuitBreaker limit-reached listener failures so check() remains non-blocking
- continue attempting later handlers after one listener throws
- add regression coverage for handler exception isolation

## Test Plan
- npm test -- CircuitBreaker.test.ts (from packages/franken-observer)
- npm run build --workspace @franken/types && npm run typecheck && npm run lint && npm run build

Closes #1180

### DIFF
--- a/packages/franken-observer/src/cost/CircuitBreaker.test.ts
+++ b/packages/franken-observer/src/cost/CircuitBreaker.test.ts
@@ -102,6 +102,32 @@ describe('CircuitBreaker', () => {
       expect(() => breaker.check(999)).not.toThrow()
     })
 
+    it('isolates handler failures and still returns the trip result', () => {
+      breaker.on('limit-reached', () => {
+        throw new Error('webhook failed')
+      })
+
+      expect(() => breaker.check(0.51)).not.toThrow()
+      expect(breaker.check(0.60)).toEqual({ tripped: true, limitUsd: 0.50, spendUsd: 0.60 })
+    })
+
+    it('attempts later handlers after an earlier limit-reached handler throws', () => {
+      const throwingHandler = vi.fn(() => {
+        throw new Error('webhook failed')
+      })
+      const laterHandler = vi.fn()
+
+      breaker.on('limit-reached', throwingHandler)
+      breaker.on('limit-reached', laterHandler)
+
+      expect(() => breaker.check(0.51)).not.toThrow()
+      expect(throwingHandler).toHaveBeenCalledOnce()
+      expect(laterHandler).toHaveBeenCalledOnce()
+      expect(laterHandler).toHaveBeenCalledWith(
+        expect.objectContaining({ tripped: true, limitUsd: 0.50, spendUsd: 0.51 }),
+      )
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1618, opened 2026-07-10
fix(observer): tighten W3C traceparent validation

### LINKED ISSUE #1073: fix(observer): tighten W3C traceparent validation
## Description
`parseTraceparent` and `formatTraceparent` do not fully enforce the W3C traceparent validity rules. The parser accepts the forbidden `ff` version and accepts extra fields for version `00`, while the formatter can emit all-zero trace/span IDs even though the parser rejects those IDs.

## Affected locations
- `packages/franken-observer/src/propagation/W3CTraceContext.ts:31-35` defines zero-ID sentinels and lower-hex regexes.
- `packages/franken-observer/src/propagation/W3CTraceContext.ts:51-65` only checks `parts.length < 4`, destructures without validating version, and accepts `ff` as shown by the current test expectation.
- `packages/franken-observer/src/propagation/W3CTraceContext.ts:81-89` validates only regex length/lowercase for formatter inputs, not all-zero IDs.
- `packages/franken-observer/src/propagation/W3CTraceContext.test.ts:43-48` explicitly treats `ff-...` as accepted future-version input, which conflicts with the spec's forbidden version value.

1. Call `parseTraceparent('ff-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01')`.
2. The current parser returns a context because it ignores the version field.
3. Call `parseTraceparent('00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01-extra')`.
4. The current parser accepts it because it only rejects fewer than four fields, even though version `00` must have exactly four fields.
5. Call `formatTraceparent({ traceId: '0'.repeat(32), parentSpanId: '0'.repeat(16), sampled: false })`.
6. The current formatter emits a header that `parseTraceparent` would reject.

## Impact
Trace propagation can accept or emit malformed headers. That makes interoperability with standards-compliant tracing tools brittle and can hide bad upstream propagation data during QA.

## Recommendation
Validate version syntax separately, reject `ff`, require exactly four fields for version `00`, and make `formatTraceparent` reject all-zero trace IDs and parent span IDs just like `parseTraceparent` does. Add regression tests for each case.

## Acceptance criteria
- `parseTraceparent` returns `null` for version `ff`.
- `parseTraceparent` returns `null` for version `00` headers with extra fields.
- `formatTraceparent` throws for all-zero trace IDs and parent span IDs.
- Existing valid sampled/unsampled headers still parse and format successfully.

Duplicate check: local issue cache and targeted GitHub search for `traceparent W3C version ff extra fields`, `w3c trace context`, `all-zero traceparent`, and `trace context` found no open issue covering this traceparent validation gap.

djm204:
## Summary
- Reject malformed W3C traceparent versions, including forbidden `ff`.
- Require version `00` traceparent headers to have exactly four fields.
- Reject all-zero trace IDs and parent span IDs when formatting traceparent headers.

## Test Plan
- `npm run test --workspace @franken/observer -- src/propagation/W3CTraceContext.test.ts`
- `npm run test --workspace @franken/observer`
- `npm run build --workspace @franken/types && npm run typecheck --workspace @franken/observer`
- `npm run lint --workspace @franken/observer` (passes with 6 existing warnings outside touched files)

Closes #1073

### DIFF
--- a/packages/franken-observer/src/propagation/W3CTraceContext.test.ts
+++ b/packages/franken-observer/src/propagation/W3CTraceContext.test.ts
@@ -40,11 +40,23 @@ describe('parseTraceparent', () => {
     expect(parseTraceparent(`00-${TRACE_ID}-${SPAN_ID}-02`)?.sampled).toBe(false)
   })
 
-  it('accepts future version bytes (forwards compatibility)', () => {
-    const result = parseTraceparent(`ff-${TRACE_ID}-${SPAN_ID}-01`)
+  it('accepts non-ff future version bytes with extra fields for forwards compatibility', () => {
+    const result = parseTraceparent(`01-${TRACE_ID}-${SPAN_ID}-01-extra`)
     expect(result).not.toBeNull()
     expect(result?.traceId).toBe(TRACE_ID)
     expect(result?.parentSpanId).toBe(SPAN_ID)
+  })
+
+  it('returns null for forbidden ff version bytes', () => {
+    expect(parseTraceparent(`ff-${TRACE_ID}-${SPAN_ID}-01`)).toBeNull()
+  })
+
+  it('returns null for version 00 headers with extra fields', () => {
+    expect(parseTraceparent(`00-${TRACE_ID}-${SPAN_ID}-01-extra`)).toBeNull()
+  })
+
+  it('returns null for malformed version bytes', () => {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1624, opened 2026-07-11
fix(observer): repair corrupt replay blobs atomically

### LINKED ISSUE #1027: fix observer replay blob writes to repair partial refs
## Description
`ReplayContentStore.put()` uses content-addressed filenames but writes blobs directly to the final path and skips writing whenever that path already exists. If a prior write was interrupted, the corrupt partial file remains at the correct hash filename; future `put()` calls return the ref without repairing the blob, and `get()` later fails hash verification.

## Affected locations
- `packages/franken-observer/src/replay/replay-content-store.ts:15-20` — computes the hash ref, checks `existsSync(path)`, and writes the content directly to the final blob path only when the file is absent.
- `packages/franken-observer/src/replay/replay-content-store.ts:24-29` — `get()` detects hash mismatch, but by then the store has already handed out a poisoned ref and has no repair/quarantine behavior.

## Risk / impact
Replay blobs are immutable evidence for deterministic replays. A single interrupted blob write can permanently poison that content hash in the store: later attempts to save the same content skip repair because the path exists, while all replay attempts for the ref fail until an operator manually deletes the blob.

## Recommendation
Write blobs via temp file + atomic rename, and when an existing blob is present, verify its hash before returning the ref. If the existing blob is corrupt, replace it atomically or quarantine it and rewrite the expected content.

## Acceptance criteria
- Blob writes are atomic and never expose partially-written final ref files.
- `put(content)` verifies an existing ref file before returning; corrupt existing blobs are repaired or quarantined with a clear error path.
- Regression coverage simulates a corrupt existing blob at the expected hash and proves a subsequent `put(content)` does not keep returning an unusable ref.

Duplicate check: searched the local open-issue cache and live GitHub for `ReplayContentStore`, `replay blob`, `hash mismatch`, `partial write`, `content-addressed`, and `atomic blobs`; no open issue covers replay blob partial-write repair, distinct from audit verification #1014 and WebhookNotifier retry issues #971/#977.

djm204:
## Summary
- verifies existing replay blobs before returning content refs
- rewrites missing/corrupt replay blobs via same-directory temp files and atomic rename
- adds regression coverage for repairing a corrupt existing blob

## Test Plan
- npm test --workspace @franken/observer -- --run src/replay/replay-content-store.test.ts
- npm run lint --workspace @franken/observer
- npm test --workspace @franken/observer

Closes #1027

### DIFF
--- a/packages/franken-observer/src/replay/replay-content-store.test.ts
+++ b/packages/franken-observer/src/replay/replay-content-store.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { existsSync, mkdtempSync, rmSync, symlinkSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { tmpdir } from 'node:os';
 import { ReplayContentStore } from './replay-content-store.js';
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1627, opened 2026-07-11
fix(observer): clone redacted spans before export

### LINKED ISSUE #1021: fix observer SpanRedactor clone spans before downstream export
## Description
`SpanRedactor` promises that the original trace object is never mutated before export, but it sometimes forwards the original `Span` objects to the downstream adapter. `flush()` creates a new trace wrapper at `packages/franken-observer/src/redaction/SpanRedactor.ts:76-79`, but `redactSpan()` returns the original `span` whenever no field on that span changes (`lines 93-97`). With no matching rules, or with rules that match only some spans, a downstream adapter that mutates span metadata/thought blocks can mutate the caller's original trace through the shared span object.

## Affected locations
- `packages/franken-observer/src/redaction/SpanRedactor.ts:45-48` — public contract says the original trace object is never mutated.
- `packages/franken-observer/src/redaction/SpanRedactor.ts:74-80` — `flush()` passes redacted spans downstream.
- `packages/franken-observer/src/redaction/SpanRedactor.ts:93-97` — unchanged spans are returned by reference instead of cloned.
- `packages/franken-observer/src/redaction/SpanRedactor.ts:100-117` — metadata is cloned only when at least one rule exists; no-rule and no-match paths preserve original metadata references.

1. Create a `SpanRedactor` with no rules, or with a rule that does not match a given span.
2. Use an underlying adapter that appends to `span.thoughtBlocks`, writes to `span.metadata`, or otherwise normalizes spans during `flush()`.
3. Call `SpanRedactor.flush(trace)`.
4. The downstream adapter receives at least one original `Span` object, so its mutation is visible on the caller's `trace.spans` despite the non-mutation contract.

## Impact
Composing redaction with other adapters is unsafe: a telemetry adapter can accidentally alter live trace state, tests may become order-dependent, and redaction's safety contract is only true for spans that actually had a field changed.

## Recommendation
Always clone span objects before passing them downstream, and clone mutable nested structures (`metadata`, `thoughtBlocks`) even when no redaction rule matches. Add tests where an underlying adapter mutates a span to prove the caller's original trace remains unchanged for no-rule, no-match, and partial-match cases.

## Acceptance criteria
- `SpanRedactor.flush()` never passes original span object references to the wrapped adapter.
- Metadata and thought-block arrays are copied before export, including no-rule and no-match paths.
- Regression tests cover downstream mutation attempts and preserve the documented non-mutation contract.

Duplicate check: searched local issue cache and targeted GitHub issue search for `SpanRedactor`, `thoughtBlocks`, `redact`, `metadata`, and `mutable span`; reviewed recent PM slot-3 claims. No open duplicate found.

djm204:
## Summary
- Always clone SpanRedactor span wrappers before passing traces to downstream adapters
- Copy metadata and thought block arrays even when no redaction rule matches
- Add regression coverage for downstream adapter mutation in no-rule, no-match, and partial-redaction cases

## Test Plan
- npm run test --workspace @franken/observer -- src/redaction/SpanRedactor.test.ts
- npm run test --workspace @franken/observer
- npm run lint --workspace @franken/observer
- git diff --check

Closes #1021

### DIFF
--- a/packages/franken-observer/src/redaction/SpanRedactor.test.ts
+++ b/packages/franken-observer/src/redaction/SpanRedactor.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect } from 'vitest'
 import { SpanRedactor } from './SpanRedactor.js'
 import { InMemoryAdapter } from '../export/InMemoryAdapter.js'
 import type { Trace, Span } from '../core/types.js'
+import type { ExportAdapter } from '../export/ExportAdapter.js'
 
 function makeSpan(overrides: Partial<Span> = {}): Span {
   return {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1629, opened 2026-07-11
test(observer): cover BatchAdapter failed drain retention

### LINKED ISSUE #1019: fix observer BatchAdapter retain traces after failed drain
## Description
`BatchAdapter.drain()` removes every buffered trace before it knows whether the underlying adapter accepted them. The method splices `this.buffer` into `batch` at `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:82`, then awaits `Promise.all(batch.map(t => this.inner.flush(t)))` at line 83. If any inner `flush()` rejects, `drain()` propagates the error but the failed traces are already gone from `this.buffer`, so a transient exporter failure permanently drops observability data.

## Affected locations
- `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:80-84` — `drain()` clears the buffer before awaiting the downstream flushes.
- `packages/franken-observer/src/adapters/batch/BatchAdapter.ts:90-95` — `stop()` relies on `drain()`, so shutdown-time exporter failures can also discard the final buffered traces.

1. Configure `BatchAdapter` with an underlying `ExportAdapter` whose `flush()` rejects for one trace, for example a temporary HTTP/SQLite failure.
2. Call `flush()` enough times to fill the buffer or call `drain()`/`stop()` manually.
3. `drain()` splices all traces out of the buffer before awaiting the downstream writes.
4. The promise rejects, but `queryByTraceId()` and the next `drain()` cannot retry those traces because they are no longer buffered.

## Impact
A retryable exporter outage causes silent trace loss at exactly the layer meant to make exporting safer. Operators lose audit/observability records even though the failed drain is visible to callers and could be retried.

## Recommendation
Keep traces in the buffer until their downstream flush succeeds, or requeue failed traces when `inner.flush()` rejects. Add regression coverage for partial and total downstream failures so successful traces are not duplicated and failed traces remain retryable.

## Acceptance criteria
- `BatchAdapter.drain()` does not permanently remove a trace from retry visibility until the corresponding `inner.flush()` succeeds.
- A rejected downstream flush leaves failed traces available for a later `drain()` or reports an explicit non-retryable discard policy.
- Unit tests cover failed drain, partial failure, and shutdown `stop()` failure behavior.

Duplicate check: searched local issue cache and targeted GitHub issue search for `BatchAdapter`, `batch adapter`, `drain`, `flush`, `trace batch`, and `lost trace`; reviewed recent PM slot-3 claims. No open duplicate found.

djm204:
## Summary
- add regression coverage that partial BatchAdapter drain failures keep failed traces retryable without duplicating successful traces
- add shutdown coverage proving stop() failures leave final buffered traces visible for retry
- verify the existing BatchAdapter implementation already satisfies issue #1019 retention semantics

## Test Plan
- npm test --workspace packages/franken-observer -- BatchAdapter.test.ts
- npm run build --workspace @franken/types && npm run typecheck --workspace packages/franken-observer
- npm run build --workspace packages/franken-observer
- npm run lint --workspace packages/franken-observer (passes with existing warnings)

Closes #1019

### DIFF
--- a/packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts
+++ b/packages/franken-observer/src/adapters/batch/BatchAdapter.test.ts
@@ -229,6 +229,36 @@ describe('BatchAdapter — drain()', () => {
     expect(calls).toEqual(['fast-fail', 'slow-success', 'fast-fail'])
   })
 
+  it('retains failed traces for retry without duplicating successful traces after partial drain failure', async () => {
+    const inner = new InMemoryAdapter()
+    const failedOnce = new Set<string>()
+    const calls: string[] = []
+    vi.spyOn(inner, 'flush').mockImplementation(async trace => {
+      calls.push(trace.id)
+      if (trace.id === 'retry-later' && !failedOnce.has(trace.id)) {
+        failedOnce.add(trace.id)
+        throw new Error('transient exporter failure')
+      }
+      await InMemoryAdapter.prototype.flush.call(inner, trace)
+    })
+    const batch = new BatchAdapter({ adapter: inner, maxBatchSize: 100 })
+
+    const alreadyPersisted = makeTrace('already-persisted')
+    const retryLater = makeTrace('retry-later')
+
+    await batch.flush(alreadyPersisted)
+    await batch.flush(retryLater)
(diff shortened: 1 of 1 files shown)


djm204:
@codex review

PR #1631, opened 2026-07-11
fix(observer): fail audit verification for missing hashed content

### LINKED ISSUE #1014: fix observer AuditTrail.verify fail on missing hashed content
## Description
`AuditTrail.verify()` reports success when an event declares `inputHash` or `outputHash` but the caller's `contentMap` does not contain the corresponding payload. That makes the verifier unable to distinguish 'content was present and matched' from 'content was never supplied', which weakens the audit trail's integrity check.

## Affected locations
- `packages/franken-observer/src/audit-event.ts:36-41` — events can record `inputHash` and/or `outputHash`.
- `packages/franken-observer/src/audit-event.ts:91-109` — `verify()` only adds a mismatch when content exists and hashes differently; missing `${eventId}:input` / `${eventId}:output` entries are silently ignored.

## Reasoning path
1. Create an audit event with `input` and `output`, so the event records hashes.
2. Later call `AuditTrail.verify(new Map())`, or provide only one side of the content.
3. Because `content !== undefined` guards the hash comparison, no mismatch is recorded for the absent hashed payload.
4. The returned `{ valid: true }` can be misread as proof that all hashed content was verified, even though no content was checked.

## Recommendation
Treat missing content for declared hashes as a verification failure, or return a structured status that separates `missing` from `mismatched`. Tests should cover empty maps, one-sided maps, matching content, and mismatched content.

## Acceptance criteria
- `AuditTrail.verify()` fails or reports `missing` when an event has `inputHash`/`outputHash` but the content map lacks that key.
- Existing mismatch behavior still flags wrong content.
- Unit tests assert that an empty content map cannot produce a clean verification result for hashed events.

Duplicate check: searched the local issue cache and live GitHub issues for `AuditTrail`, `audit trail`, `hash mismatch`, and `missing content`; no open duplicate found.

djm204:
## Summary
- Make `AuditTrail.verify()` fail when an event declares an input/output hash but the caller omits that content from the verification map.
- Add regression coverage for missing input content, missing output content, one-sided maps, and existing match/mismatch behavior.

## Test Plan
- `npm test --workspace @franken/observer -- --run src/audit-event.test.ts`
- `npm run typecheck --workspace @franken/observer`
- `npm run lint --workspace @franken/observer` (passes with 6 pre-existing warnings)
- `npm test --workspace @franken/observer`

Closes #1014

### DIFF
--- a/packages/franken-observer/src/audit-event.test.ts
+++ b/packages/franken-observer/src/audit-event.test.ts
@@ -149,6 +149,47 @@ describe('AuditTrail', () => {
     expect(result.mismatches).toHaveLength(1);
   });
 
+  it('verify() fails when hashed input content is missing', () => {
+    const trail = new AuditTrail();
+    const event = createAuditEvent('test', {}, { phase: 'p', provider: 'pr', input: 'hello' });
+    trail.append(event);
+
+    expect(trail.verify(new Map())).toEqual({
+      valid: false,
+      mismatches: [`${event.eventId}: input content missing`],
+    });
+  });
+
+  it('verify() fails when hashed output content is missing', () => {
+    const trail = new AuditTrail();
+    const event = createAuditEvent('test', {}, { phase: 'p', provider: 'pr', output: 'result' });
+    trail.append(event);
+
+    expect(trail.verify(new Map())).toEqual({
+      valid: false,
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1639, opened 2026-07-11
fix(beasts): reject malformed daemon pid files

### LINKED ISSUE #995: fix(beasts): reject malformed daemon pid-file contents
## Description
The beasts daemon pid-file reader accepts partially numeric contents because it uses `Number.parseInt(raw.trim(), 10)` and only checks that the parsed value is finite and positive. A corrupted pid file such as `123abc` is treated as PID 123 instead of being rejected and removed.

## Affected locations
- `packages/franken-orchestrator/src/http/beast-daemon-server.ts:126-134` — `readExistingPidFile` parses `raw.trim()` with `Number.parseInt` before deciding whether to remove the file.
- `packages/franken-orchestrator/src/http/beast-daemon-server.ts:143-147` — `readPidFile` uses the same partial parse during pid-file release checks.

1. Leave a stale/corrupt daemon pid file containing a partially numeric value, for example `123abc`.
2. Start the beasts daemon while an unrelated process with PID 123 is alive.
3. `readExistingPidFile` returns `123`; `claimPidFile` then reports `beasts-daemon is already running with PID 123` instead of treating the pid file as malformed and removing it.

## Impact
A single corrupt pid-file write or manual edit can block local beast-daemon startup behind an unrelated process id. This is low-risk but operator-visible and hard to diagnose because the file contents do not actually name a valid PID.

## Recommendation
Validate pid-file contents with a full decimal-integer match such as `/^\d+$/` before converting, and apply the same canonical validation in both readers. Malformed contents should be removed in `readExistingPidFile` and treated as absent in `readPidFile`.

## Acceptance criteria
- `123abc`, `123\n456`, blank, negative, and decimal pid-file contents are rejected rather than partially parsed.
- A malformed existing pid file is removed and does not block daemon startup.
- Unit coverage exercises both `readExistingPidFile`/claim behavior and release behavior.

Duplicate check: searched the open issue cache and targeted GitHub queries for `pid file`, `beasts-daemon`, `daemon pid`, and `parseInt`; no open duplicate was found.

djm204:
## Summary
- Add strict full-decimal validation for beasts daemon pid-file contents before converting to a number
- Treat malformed existing pid files as stale so daemon startup can reclaim them
- Treat malformed release-time pid files as absent so shutdown removes them

## Test Plan
- npm --workspace @franken/orchestrator test -- tests/integration/beasts/beast-daemon.test.ts
- npm --workspace @franken/orchestrator run lint
- npm run build
- npm --workspace @franken/orchestrator run typecheck

Closes #995

### DIFF
--- a/packages/franken-orchestrator/src/http/beast-daemon-server.ts
+++ b/packages/franken-orchestrator/src/http/beast-daemon-server.ts
@@ -49,6 +49,7 @@ function errorMessage(error: unknown): string {
 const DEFAULT_HOST = '127.0.0.1';
 const DEFAULT_PORT = 4050;
 const TERMINAL_RUN_STATUSES = new Set(['completed', 'failed', 'stopped']);
+const PID_FILE_DECIMAL_PATTERN = /^\d+$/;
 
 export function defaultBeastDaemonPidFile(root: string): string {
   return join(root, '.frankenbeast', 'beasts-daemon.pid');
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #1658, opened 2026-07-11
test(vitest): parse suite env flags strictly

### LINKED ISSUE #1038: test: parse Vitest suite env flags strictly
## Description
Vitest suite selection treats any non-empty `E2E`, `INTEGRATION`, or `EVAL` environment value as enabled, so common false-like values such as `E2E=false`, `INTEGRATION=0`, or `EVAL=false` can select the opt-in suites instead of the default/unit suite.

## Affected locations
- `packages/franken-orchestrator/vitest.config.ts:6-14` computes `isIntegration` and `isE2e` with `Boolean(process.env['INTEGRATION'])` / `Boolean(process.env['E2E'])`, then uses those booleans to choose `tests/integration/**` or `tests/e2e/**` includes.
- `packages/franken-observer/vitest.config.ts:3-20` computes `isIntegration` and `isEval` with `Boolean(process.env['INTEGRATION'])` / `Boolean(process.env['EVAL'])`. Combined with `passWithNoTests: true`, `EVAL=false npm run test` can select the eval glob and still pass even when no eval tests exist.
- `packages/franken-orchestrator/tests/e2e/chunk-pipeline.test.ts:27` and `tests/e2e/cli-skill-execution.test.ts:29` also use `describe.skipIf(!process.env['E2E'])`, so direct path runs can execute E2E tests when `E2E=false` is present.

1. In a shell or CI job that exports `E2E=false` or `INTEGRATION=false`, run the orchestrator package's default Vitest command.
2. The config sees the string `"false"` as truthy, flips `runE2e`/`runIntegration` on, and changes the include set away from unit tests.
3. For observer eval mode, `EVAL=false` selects `src/**/*.eval.test.ts`; because there are currently no eval tests and `passWithNoTests` is enabled, the command can report success while running no useful assertions.

## Impact
This makes test/CI reproduction misleading: a developer or CI wrapper that passes explicit false-like env values can silently run the wrong suite, skip the intended unit suite, or get a green no-test eval run. That undermines regression coverage and makes opt-in E2E/eval behavior harder to reason about.

## Recommendation
Use a shared strict env parser for Vitest suite flags, e.g. only enable when the value is `"1"` or `"true"` (case-insensitive), treat `"0"`, `"false"`, `"no"`, and empty values as disabled, and update both config-level include selection and `describe.skipIf` guards to use it. Add regression tests or a config test that asserts `E2E=false`, `INTEGRATION=false`, and `EVAL=false` do not select opt-in suites.

## Acceptance criteria
- `E2E=false` and `INTEGRATION=false` keep orchestrator on the default/unit include set unless explicit test paths are requested.
- `EVAL=false` keeps observer on the default unit include set and cannot produce a green no-test eval run.
- E2E test guards use the same strict parser as the Vitest config.
- Positive opt-ins such as `E2E=true`, `INTEGRATION=true`, or `EVAL=true` continue to work.

Duplicate check: read the local open-issue cache and searched live GitHub for `E2E=false`, `INTEGRATION=false`, `Boolean process.env`, and `vitest env flag`; reviewed related #966 (`test-observer-remove-passWithNoTests-mask`) and #945 (`test add root level integration and eval test entrypoints`) and found they cover no-test masking/discoverability, not false-like env flag parsing across Vitest configs and E2E guards.

djm204:
## Summary
- route E2E suite gating through `readVitestFlag` instead of direct `process.env.E2E` checks
- add regression coverage that false-like suite flags keep default unit test discovery active
- strengthen orchestrator and observer Vitest config/discovery tests around strict suite flag parsing

## Test Plan
- [x] `npm exec -- vitest run tests/unit/vitest-env.test.ts` (from `packages/franken-orchestrator`)
- [x] `npm exec -- vitest run src/vitest-discovery.test.ts` (from `packages/franken-observer`)
- [x] dependency builds + typecheck/lint for orchestrator/observer related packages
- [x] `npm run test --workspace @franken/orchestrator`
- [x] `npm run test --workspace @franken/observer`

Closes #1038

### DIFF
--- a/packages/franken-observer/src/vitest-discovery.test.ts
+++ b/packages/franken-observer/src/vitest-discovery.test.ts
@@ -2,7 +2,14 @@ import { spawnSync } from 'node:child_process';
 import { readFileSync } from 'node:fs';
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
+
+type VitestConfig = {
+  test?: {
+    include?: string[];
+    exclude?: string[];
+  };
+};
 
 const packageRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 const discoverySmokeTimeoutMs = 20_000;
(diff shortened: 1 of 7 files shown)


djm204:
@codex review

djm204:
@codex review

PR #1943, opened 2026-07-12
test(orchestrator): stabilize http abort disconnect test

### LINKED ISSUE #1932: test(franken-orchestrator): avoid flaky disconnect timing in http server utils unit test
The abort-signal unit test in http-server-utils.test uses setTimeout to destroy the client request after 10 ms. On slow CI hosts this can race before the request handler attaches, causing non deterministic failures. Consider waiting longer or using an explicit server side signal before destroying the request so test deterministically triggers abort handling and avoids CI flakiness.

djm204:
## Summary
- keep the http-server-utils abort-disconnect test cleanup in a finally block
- document the handler-ready synchronization so the test remains independent of wall-clock disconnect timing

## Test plan
- npm run test --workspace @franken/orchestrator -- tests/unit/http/http-server-utils.test.ts
- npm run build
- npm run typecheck --workspace @franken/orchestrator
- npm run lint --workspace @franken/orchestrator (passes with existing warnings)

Closes #1932

### DIFF
--- a/packages/franken-orchestrator/tests/unit/http/http-server-utils.test.ts
+++ b/packages/franken-orchestrator/tests/unit/http/http-server-utils.test.ts
@@ -32,18 +32,24 @@ describe('http-server-utils', () => {
       throw new Error('server did not bind to a TCP address');
     }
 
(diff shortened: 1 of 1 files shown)


djm204:
@codex review

PR #2001, opened 2026-07-12
fix(mcp-suite): validate memory query limit safety

### LINKED ISSUE #1975: fix(mcp-suite): validate fbeast_memory_query limit values
## Description
`fbeast_memory_query` accepts the optional `limit` argument as a string in its MCP schema, but the handler converts it with `Number(args['limit'])` and forwards the result without validation. Non-finite, zero, negative, or non-integer values can change the query result shape instead of falling back to a safe default or returning a validation error.

## Affected locations
- `packages/franken-mcp-suite/src/servers/memory.ts:19-32` — schema declares `limit` as a string and the handler uses `Number(args['limit'])` before calling `brain.query(...)`.
- `packages/franken-mcp-suite/src/adapters/brain-adapter.ts:52-80` — `input.limit` is forwarded to `brain.episodic.recall(...)` and then reused in `results.slice(0, input.limit ?? 20)`.
- `packages/franken-brain/src/sqlite-brain.ts:232-267` and `:280-284` — episodic recall/recent bind the caller-provided value into SQL `LIMIT ?`.
- `packages/franken-mcp-suite/src/servers/memory.test.ts:24-47` — current coverage only verifies a valid numeric limit path and does not cover invalid strings, `NaN`, negative values, zero, or fractional values.

1. Call `fbeast_memory_query` with a required `query` and an invalid limit such as `"abc"`, `"NaN"`, `"-1"`, or `"0"`.
2. `memory.ts` computes `Number(args['limit'])` and passes `NaN`, `-1`, or `0` as the limit.
3. The adapter forwards that value to episodic SQL `LIMIT ?` and also applies `results.slice(0, input.limit ?? 20)` to merged results. Invalid, negative, or zero values can make the MCP response empty, unexpectedly shorten result sets, or fail at the SQLite binding layer rather than behaving like the documented default max-results option.

## Impact
This is a low-risk reliability bug in the MCP memory server. A malformed tool argument can make memory look empty, return the wrong number of rows, or surface an internal query/binding failure. That is especially confusing because the tool schema documents `limit` as an optional max-results value with default 20, but does not communicate that malformed strings can change control flow.

## Recommendation
Validate `limit` at the MCP boundary before calling `brain.query`:
- parse only finite positive integers;
- reject non-numeric, zero, negative, fractional, or unsafe values with an MCP `isError` validation response, or clamp to a documented safe range;
- keep the documented default of 20 when `limit` is omitted;
- add unit tests for omitted, valid, invalid string, zero, negative, fractional, and very large limits.

## Acceptance criteria
- Invalid `limit` inputs cannot reach `brain.query`, `SqliteEpisodicMemory.recall`, `recent`, or `Array.prototype.slice` as `NaN`, negative, zero, fractional, or unsafe numbers.
- `fbeast_memory_query` returns a clear validation error or documented clamp behavior for malformed limits.
- Unit tests cover the invalid limit cases and the valid/default paths.

Duplicate check: read `/home/pfkagent/.hermes/state/frankenbeast_open_issues_cache.json` first and checked terms `memory`, `fbeast_memory_query`, `limit`, `NaN`, `MCP`, `franken-mcp`, and `query limit`; targeted `gh issue list --search "fbeast_memory_query OR \"memory query\" OR \"MCP memory\" OR \"query limit\""` returned no open issue for this MCP memory-query limit path. Related open issues #1958, #1922, #1742, and #607 cover different adapters/rate limits/security parsing, not this tool argument validation bug.

djm204:
## Summary
- Require `fbeast_memory_query` limits to be safe integers before calling the brain adapter
- Extend memory server coverage for fractional, out-of-range, and unsafe limit strings

Closes #1975

## Test plan
- `npm run test --workspace @franken/mcp-suite -- src/servers/memory.test.ts`

## Notes
- fbeast MCP tools were not available in this worker schema, so verification used normal terminal/git/gh tooling.

### DIFF
--- a/packages/franken-mcp-suite/src/servers/memory.test.ts
+++ b/packages/franken-mcp-suite/src/servers/memory.test.ts
@@ -91,7 +91,7 @@ describe('Memory Server', () => {
   });
 
   it('rejects invalid query limits before calling the brain adapter', async () => {
-    for (const invalidLimit of ['abc', 'NaN', 'Infinity', '0', '-1']) {
+    for (const invalidLimit of ['abc', 'NaN', 'Infinity', '0', '-1', '1.5', '1001', '9007199254740993']) {
       const brain = {
         query: vi.fn().mockResolvedValue([]),
         store: vi.fn(),
--- a/packages/franken-mcp-suite/src/shared/tool-registry.ts
+++ b/packages/franken-mcp-suite/src/shared/tool-registry.ts
@@ -51,7 +51,7 @@ function parseMemoryQueryLimit(value: unknown): { ok: true; value: number } | { 
     return { ok: false, message: `limit must be a positive integer between 1 and ${MAX_MEMORY_QUERY_LIMIT}` };
   }
   const parsed = Number(raw);
-  if (!Number.isFinite(parsed) || !Number.isInteger(parsed) || parsed < 1 || parsed > MAX_MEMORY_QUERY_LIMIT) {
+  if (!Number.isFinite(parsed) || !Number.isSafeInteger(parsed) || parsed < 1 || parsed > MAX_MEMORY_QUERY_LIMIT) {
     return { ok: false, message: `limit must be a positive integer between 1 and ${MAX_MEMORY_QUERY_LIMIT}` };
   }
   return { ok: true, value: parsed };


djm204:
@codex review

PR #2002, opened 2026-07-12
fix(governor): reject duplicate approval waiters

### LINKED ISSUE #1956: fix(governor): reject duplicate approval waiter requestIds
## Description
`ApprovalWaiterRegistry.waitFor()` unconditionally writes `pending.set(requestId, ...)` when a real waiter registers. If two callers accidentally use the same `requestId`, the second waiter replaces the first resolver. A later `resolve(requestId, ...)` only wakes the second promise and deletes the entry, leaving the first caller permanently pending until some outer timeout fires.

## Affected locations
- `packages/franken-governor/src/gateway/approval-waiter-registry.ts:49-65` — `register()` preserves an existing resolver, but `waitFor()` overwrites any existing entry for the same `requestId`.
- `packages/franken-governor/src/gateway/approval-waiter-registry.ts:73-78` — `resolve()` deletes the single surviving entry and cannot notify an overwritten waiter.
- `packages/franken-governor/tests/unit/server/http-approval-waiters.test.ts:31-160` covers the happy path, signature propagation, and timeout cleanup, but there is no collision regression for duplicate `requestId` waiters.

1. Create an `ApprovalWaiterRegistry`.
2. Call `waitFor('req-1', ...)` and keep promise A.
3. Call `waitFor('req-1', ...)` again and keep promise B.
4. Call `resolve('req-1', { requestId: 'req-1', decision: 'APPROVE', ... })`.
5. Promise B resolves, but promise A has no resolver left in the registry and remains pending until an unrelated timeout path handles it.

## Impact
Approval flows that accidentally reuse a request id can strand the original caller, causing confusing timeouts even though the operator response was accepted. The current API exposes no signal that a duplicate waiter replaced an active one.

## Recommendation
Make `waitFor()` fail fast when `requestId` already has a real waiter, or explicitly support fan-out by storing multiple resolvers per request id. Add a regression test that a duplicate waiter cannot silently orphan the first promise.

## Acceptance criteria
- Duplicate `waitFor()` calls for the same `requestId` are rejected or all waiters are resolved deterministically.
- `register()` placeholder behavior still preserves real waiters.
- A unit test covers duplicate `requestId` registration and proves no pending promise is orphaned.

Duplicate check: local cache plus targeted `gh issue list` searches for `ApprovalWaiterRegistry duplicate requestId waitFor` and `approval waiter registry duplicate` returned no open matches.

djm204:
## Summary
- track whether an approval registry entry has a real waiter versus a placeholder
- reject duplicate real waiters for the same requestId instead of overwriting the original resolver
- cover duplicate waiter rejection and placeholder refresh preservation with regression tests

## Test Plan
- npm test --workspace @franken/governor -- tests/unit/server/http-approval-waiters.test.ts
- npm run lint --workspace @franken/governor
- npm run build --workspace @franken/types && npm run typecheck --workspace @franken/governor && npm run build --workspace @franken/governor

Closes #1956

### DIFF
--- a/packages/franken-governor/src/gateway/approval-waiter-registry.ts
+++ b/packages/franken-governor/src/gateway/approval-waiter-registry.ts
@@ -3,6 +3,7 @@ import type { ApprovalResponse } from '../core/types.js';
 interface PendingApproval {
   readonly taskId: string;
   readonly summary: string;
+  readonly hasRealWaiter: boolean;
   resolve: (response: ApprovalResponse) => void;
 }
 
@@ -51,6 +52,7 @@ export class ApprovalWaiterRegistry {
     this.pending.set(requestId, {
       taskId,
       summary,
+      hasRealWaiter: existing?.hasRealWaiter ?? false,
       resolve: existing?.resolve ?? (() => {}),
     });
   }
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2005, opened 2026-07-12
fix(governor): sweep expired session tokens

### LINKED ISSUE #1966: fix(governor): sweep expired SessionTokenStore entries
## Bug Description
`SessionTokenStore` only deletes an expired token when that exact `tokenId` is read. Expired tokens that are never queried remain in the private `Map` indefinitely because the store has no `sweepExpired()`, max-size policy, insertion-time cleanup, or clock-injected compaction path.

## Affected locations
- `packages/franken-governor/src/security/session-token-store.ts:3-8` stores every token in a process-local `Map`.
- `packages/franken-governor/src/security/session-token-store.ts:10-17` removes expired entries only on `get(tokenId)` for that same id.
- `packages/franken-governor/src/security/session-token-store.ts:22-28` exposes `revoke()`/`isValid()` but no global cleanup path.
- `packages/franken-governor/tests/unit/security/session-token-store.test.ts:37-74` verifies lookup-time expiry for one token but does not cover stale unqueried tokens.

## Reasoning path
1. Approval flows create session tokens with `createSessionToken()` and store them in `SessionTokenStore`.
2. If many approval tokens expire naturally and callers never check those old ids again, the only expiry deletion branch is never reached.
3. Long-running governor processes therefore retain expired session-token objects for the lifetime of the process.

## Impact
A long-lived approval service can accumulate stale session-token state, making memory use and health/debug output less predictable. The current tests prove expired tokens are hidden when queried, but they do not prove the store is bounded or cleaned up when tokens are abandoned.

## Recommendation
Add an explicit cleanup mechanism such as `sweepExpired(now?)`, opportunistic sweeping before/after `store()`, or a bounded retention policy. Prefer injecting `now`/clock for deterministic tests rather than relying only on global `Date.now()`.

## Acceptance Criteria
- [ ] Expired tokens can be swept even when their ids are never queried again.
- [ ] Tests cover multiple expired/unexpired tokens and assert that only expired entries are removed.
- [ ] Existing `get()`, `isValid()`, and `revoke()` behavior remains unchanged for valid and explicitly revoked tokens.
- [ ] The implementation provides a deterministic testing hook or clock strategy for expiry behavior.

Duplicate check: local issue cache and targeted GitHub searches for `SessionTokenStore expired tokens`, `session token cleanup`, `token store TTL`, and `expired token` returned no existing open issue covering abandoned expired-token retention.

djm204:
## Summary
- Add public `SessionTokenStore.sweepExpired({ nowMs })` for explicit cleanup of abandoned expired tokens.
- Keep `cleanupExpired()` as a compatibility alias.
- Export the sweep options type and cover in-memory plus persisted sweep behavior.

Closes #1966

## Test plan
- `npm exec --workspace @franken/governor -- vitest run tests/unit/security/session-token-store.test.ts`
- `npm run typecheck --workspace @franken/governor`

### DIFF
--- a/packages/franken-governor/src/index.ts
+++ b/packages/franken-governor/src/index.ts
@@ -61,7 +61,11 @@ export { GovernorAuditRecorder } from './audit/index.js';
 export { formatApprovalResponseSignaturePayload, SignatureVerifier } from './security/index.js';
 export type { ApprovalResponseSignaturePayloadFields } from './security/index.js';
 export { createSessionToken, SessionTokenStore } from './security/index.js';
-export type { CreateSessionTokenParams, SessionTokenStoreOptions } from './security/index.js';
+export type {
+  CreateSessionTokenParams,
+  SessionTokenStoreOptions,
+  SweepExpiredSessionTokenOptions,
+} from './security/index.js';
 
 export { CliChannel } from './channels/index.js';
 export type { ReadlineAdapter, CliChannelDeps } from './channels/index.js';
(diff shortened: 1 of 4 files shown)


djm204:
@codex review

PR #2091, opened 2026-07-13
fix(web): reconnect Beast SSE after malformed payloads

### LINKED ISSUE #2082: fix(web): reconnect Beast SSE after malformed event payloads
## Description
`BeastApiClient.subscribeToEvents()` records malformed SSE payloads as a failed `lastEventId`, but it only reports the parse error and keeps the existing `EventSource` open. Subsequent valid events are delivered to handlers, yet `rememberProcessedEventId()` refuses to advance `lastEventId` while `failedEventId` is set to the earlier bad event. The next network reconnect therefore resumes from the old cursor and can replay already-handled run/agent events.

## Affected locations
- `packages/franken-web/src/lib/beast-api.ts:181-196` keeps `lastEventId` and `failedEventId`; once `failedEventId` is set, `rememberProcessedEventId()` ignores later event ids until the exact failed id is processed again.
- `packages/franken-web/src/lib/beast-api.ts:207-215` calls `rememberFailedEventId(event)` on JSON/parse failure, rethrows, and the per-event listeners only call `handlers.error`.
- `packages/franken-web/src/lib/beast-api.ts:241-282` catch malformed `snapshot`, `agent.status`, `agent.event`, `run.status`, `run.log`, and `run.event` payloads but do not close the stream or schedule a reconnect, so the failed event is never replayed while the stream remains alive.
- Existing tests cover dashboard malformed snapshots (`packages/franken-web/src/lib/dashboard-api.test.ts:215`) but `git grep` found no `subscribeToEvents` regression around malformed Beast SSE payloads, `failedEventId`, or `lastEventId` replay behavior.

1. A Beast SSE stream receives event id `10` with malformed JSON in a `run.log` event.
2. `handleEvent()` catches the parse failure, sets `failedEventId = "10"`, and reports the error.
3. The stream stays open and then receives valid event id `11`; the handler runs, but `rememberProcessedEventId()` returns early because `failedEventId` is still `"10"` and event id `11` does not match it.
4. On a later reconnect, the client sends the old `lastEventId` instead of `11`, so the server can replay events the UI already handled after the malformed payload.

## Impact
A single malformed or schema-incompatible Beast SSE event can leave the client cursor stale until the next transport reconnect. Operators may see duplicate log lines/status transitions or re-applied run events after any later network blip, which makes the Beasts dashboard less trustworthy during long-running sessions.

## Recommendation
When an event with `lastEventId` fails to parse, either close the current `EventSource` and schedule a reconnect from the previous safe cursor so the failed event can be retried in order, or explicitly advance/drop the malformed event with a recorded error so later valid event ids can move `lastEventId` forward. Add a regression test for malformed Beast SSE payloads followed by a valid event and a reconnect URL assertion.

## Acceptance criteria
- Malformed Beast SSE payloads do not leave `failedEventId` permanently blocking `lastEventId` advancement.
- A valid event after a malformed event is not replayed on the next reconnect unless the client intentionally retries from the failed event.
- Tests cover at least one malformed `run.log` or `agent.event` payload followed by a valid event and reconnect.

Duplicate check: local issue cache searched for `lastEventId`, `beast event`, `event stream`, `sse`, `malformed snapshot`, `run.log`, and `run event`. Targeted live issue #1699 covers a persistent reconnect error banner after successful SSE resume, but not malformed payload cursor freeze/replay behavior.

djm204:
## Summary
- close and reconnect the Beast SSE stream when a handled event payload fails to parse
- keep the reconnect cursor at the last successfully parsed event id and ignore stale events from the closed source
- add a regression covering malformed run.log payloads followed by later source events and reconnect URL cursor behavior

## Test Plan
- npm test --workspace @franken/web -- --run tests/lib/beast-api.test.ts -t 'reconnects after malformed handled SSE payloads'
- npm test --workspace @franken/web -- --run tests/lib/beast-api.test.ts
- npm test --workspace @franken/web
- npm run build --workspace @franken/types && npm run typecheck --workspace @franken/web
- npm run lint --workspace @franken/web
- npm run build --workspace @franken/web

Closes #2082

### DIFF
--- a/packages/franken-web/src/lib/beast-api.ts
+++ b/packages/franken-web/src/lib/beast-api.ts
@@ -180,6 +180,12 @@ export class BeastApiClient {
     let reconnectTimer: ReturnType<typeof setTimeout> | undefined;
     let lastEventId: string | undefined;
     let failedEventId: string | undefined;
+    class MalformedSsePayloadError extends Error {
+      constructor(readonly originalError: unknown) {
+        super(toError(originalError).message);
+        this.name = 'MalformedSsePayloadError';
+      }
+    }
     const parse = <T>(event: MessageEvent): T => JSON.parse(event.data) as T;
     const parseWithEventId = <T extends object>(event: MessageEvent): T & { eventId?: string } => {
       const parsed = parse<T>(event);
@@ -209,7 +215,7 @@ export class BeastApiClient {
         payload = parsePayload(event);
       } catch (error) {
         rememberFailedEventId(event);
-        throw error;
+        throw new MalformedSsePayloadError(error);
       }
       rememberProcessedEventId(event);
       handler(payload);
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2142, opened 2026-07-13
fix(ci): preserve component latest without root release

### LINKED ISSUE #2141: ci: avoid removing latest release with no root tag
Description: component-only latest release can be demoted with no replacement when no root tag exists, leaving no latest release. Recommendation: guard zero-root case and add regression test.

djm204:
Fixes #2141.

## Summary
- Look up the most recent root release before demoting a component release marked latest.
- Exit without demoting when no root tag exists so GitHub retains the current latest release.
- Add a release workflow regression assertion for the guard ordering.

## Test plan
- `corepack npm@11.5.1 run test:root -- tests/unit/ci-workflow.test.ts`
- `git diff --check`

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -73,16 +73,20 @@ jobs:
             exit 0
           fi
 
-          # Demote the component release
+          # Promote the most recent root release back to latest. If no root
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2152, opened 2026-07-13
test(ci): tighten Node version assertions

### LINKED ISSUE #2140: test(ci): tighten CI Node version assertion
Current CI workflow test uses a loose node-version match and can pass on malformed values.

djm204:
Closes #2140

## Summary
- centralize CI setup-node Node-version assertions behind a stricter helper
- reject malformed setup-node node-version-like inputs while allowing only node-version-file: .nvmrc
- assert .nvmrc uses an exact semver pin before checking the expected baseline

## Test plan
- npm run test:root -- tests/unit/ci-workflow.test.ts

Note: fbeast MCP tools were not available in this worker, so I followed AGENTS.md with normal git/GitHub/terminal verification.


djm204:
@codex review

PR #2162, opened 2026-07-13
fix(observer): validate webhook retry bounds

### LINKED ISSUE #2156: fix(observer): validate WebhookNotifier retry option bounds
## Description
Closed issue #2015 says WebhookNotifier should validate invalid retry configuration, but the current implementation still copies caller-provided retry values directly and computes attempts/delays from them. A bad `maxRetries`, `baseDelayMs`, or `maxDelayMs` can skip the initial webhook attempt or send nonsensical delay values to the sleep function.

## Affected locations
- `packages/franken-observer/src/notify/WebhookNotifier.ts:72-78` stores `maxRetries`, `baseDelayMs`, `maxDelayMs`, and `jitter` without range/finite/integer validation.
- `packages/franken-observer/src/notify/WebhookNotifier.ts:82-91` computes `maxAttempts = 1 + maxRetries` and passes computed delays directly to `sleepFn`.
- `packages/franken-observer/src/notify/WebhookNotifier.test.ts:134-247` covers normal retry/backoff behavior but not invalid retry option values.

1. Construct `new WebhookNotifier({ url, fetch, retry: { maxRetries: -1 } })`.
2. Call `send(payload)`.
3. `maxAttempts` becomes `0`, so the retry loop never calls `fetchFn`; `lastError` remains undefined and `send()` throws/rejects with undefined instead of making the required initial delivery attempt or reporting a clear config error.
4. Similar invalid values such as `NaN`, negative `baseDelayMs`, or negative/non-finite `maxDelayMs` can produce NaN/negative/Infinity sleep delays on retry paths.

## Impact
Low-risk reliability/maintainability issue: a misconfigured notifier can silently suppress every webhook attempt or fail with an unhelpful non-Error rejection. Because this path is used for fire-and-forget HITL/circuit-breaker style notifications, bad config can make delivery failures hard to diagnose.

## Recommendation
Normalize and validate retry options in the constructor or a helper before `send()` uses them:
- require `maxRetries` to be a finite non-negative integer, or clamp/floor it with documented behavior;
- require `baseDelayMs` and `maxDelayMs` to be finite non-negative numbers;
- ensure jittered delay values remain finite, non-negative, and capped by `maxDelayMs`;
- throw a clear configuration error instead of allowing the attempt loop to skip the first delivery.

## Acceptance criteria
- Regression tests cover negative/NaN `maxRetries` and prove `WebhookNotifier.send()` never skips the initial attempt silently.
- Invalid delay bounds either throw clear validation errors or are explicitly clamped to documented safe values.
- Delays passed to `sleepFn` are finite, non-negative, and bounded for valid retry settings.
- Existing successful retry and non-retry behavior remains unchanged.

Duplicate check: Read `/home/pfkagent/.hermes/state/frankenbeast_open_issues_cache.json` and searched live GitHub for `WebhookNotifier`, `webhook retry`, `maxRetries`, `baseDelayMs`, and `maxDelayMs`. Open #2019 covers retrying non-retriable webhook responses, not invalid option validation. Closed #2015 is the exact stale-closeout predecessor; closed #1148 covers `fetchWithRetry`, and closed #977/#971 cover jitter/non-transient semantics rather than this current invalid-config failure.

djm204:
## Summary
- validate WebhookNotifier retry delay bounds during construction
- add regression coverage for invalid baseDelayMs/maxDelayMs and bounded finite sleep delays

Closes #2156

## Test plan
- `npm run test --workspace @franken/observer -- WebhookNotifier.test.ts`
- `git diff --check`

### DIFF
--- a/packages/franken-observer/src/notify/WebhookNotifier.test.ts
+++ b/packages/franken-observer/src/notify/WebhookNotifier.test.ts
@@ -370,6 +370,39 @@ describe('WebhookNotifier', () => {
       expect(mockFetch).not.toHaveBeenCalled()
     })
 
+    it('rejects invalid retry delay bounds during retry configuration validation', () => {
+      const invalidOptions = [
+        { retry: { maxRetries: 1, baseDelayMs: -1 } },
+        { retry: { maxRetries: 1, baseDelayMs: Number.NaN } },
+        { retry: { maxRetries: 1, baseDelayMs: Number.POSITIVE_INFINITY } },
+        { retry: { maxRetries: 1, maxDelayMs: -1 } },
+        { retry: { maxRetries: 1, maxDelayMs: Number.NaN } },
+        { retry: { maxRetries: 1, maxDelayMs: Number.POSITIVE_INFINITY } },
+      ]
+
+      for (const options of invalidOptions) {
+        expect(() => createNotifier(options)).toThrow(/retry\.(baseDelayMs|maxDelayMs) must be a finite non-negative number/)
+      }
+      expect(mockFetch).not.toHaveBeenCalled()
+    })
+
+    it('passes only finite non-negative bounded delays to sleepFn', async () => {
+      mockFetch.mockResolvedValue({ ok: false, status: 503, statusText: 'Service Unavailable' })
+      const sleepFn = vi.fn().mockResolvedValue(undefined)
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2168, opened 2026-07-13
fix(live-bench): bound long model path segments

### LINKED ISSUE #2165: fix(live-bench): bound model path segments for long model ids
## Description
`WorkspaceProvisioner` encodes the full model id into a single directory component by writing each UTF-16 code unit as hex and prefixing it with `model-` (`packages/live-bench/src/workspace/workspace-provisioner.ts:159-165`). On this checkout the filesystem `NAME_MAX` is 255 bytes, so any model id longer than 125 UTF-16 code units produces a component longer than 255 characters. Provisioning then fails while creating `runDir`/`workspaceDir` instead of recording a benchmark row.

This is reachable from normal benchmark matrix input because `BenchmarkMatrixRow.model` is an unconstrained string (`packages/live-bench/src/types.ts:27-38`). Current tests cover punctuation, lone surrogates, and case-folding safety (`packages/live-bench/tests/workspace-provisioner.test.ts:123-163`) but do not cover long model ids or the filesystem component limit.

## Affected locations
- `packages/live-bench/src/workspace/workspace-provisioner.ts:54` derives `modelSegment` directly from `row.model`.
- `packages/live-bench/src/workspace/workspace-provisioner.ts:67-70` uses that segment as one path component under `runDir`.
- `packages/live-bench/src/workspace/workspace-provisioner.ts:159-165` expands every model UTF-16 code unit to four hex characters without hashing/truncation.
- `packages/live-bench/tests/workspace-provisioner.test.ts:123-163` verifies collision/case behavior but misses long-model bounds.

1. Use a live-bench row whose `model` is a long provider id or synthetic model name of 126 ASCII characters.
2. `modelPathSegment()` returns `model-` plus 252 hex characters, i.e. 258 characters in one path component.
3. On a common filesystem with `NAME_MAX=255`, the `mkdirSync(workspaceDir, { recursive: true })` path creation fails with an overlong component instead of provisioning the benchmark workspace.

## Impact
Low-risk reliability bug: benchmark runs for long model ids can fail before the client starts, making the benchmark matrix brittle and producing no environment metadata/evidence for that row.

## Recommendation
Keep the collision/case-folding guarantees but bound the path component length. For example, use a short safe prefix plus a cryptographic digest of the full model id (and optionally a truncated readable slug), while preserving the original model string in `environment.json`.

## Acceptance criteria
- A long `row.model` value over the filesystem component limit provisions successfully.
- Distinct long model ids still produce distinct `runDir` values.
- Existing punctuation, lone-surrogate, and case-folding tests continue to pass.
- `environment.json` still records the original full model id.

Duplicate check: local open issue cache terms checked: `live-bench`, `workspace`, `modelPathSegment`, `model path length`, `ENAMETOOLONG`, `runTimestamp`; targeted all-state GitHub search `modelPathSegment OR live-bench model path length OR ENAMETOOLONG` returned no matches.

djm204:
## Summary
- bound long live-bench model path segments with a sha256 digest once the encoded component would exceed filesystem NAME_MAX
- preserve existing full UTF-16 hex segments for short model ids
- add regression coverage for long model ids, distinct directories, successful provisioning, and environment metadata preservation

Closes #2165

## Test plan
- `npm run build --workspace @franken/types`
- `npm run test --workspace @franken/live-bench -- --run tests/workspace-provisioner.test.ts`
- `npm run typecheck --workspace @franken/live-bench`

Note: this isolated worktree initially had no node_modules, so targeted package dependencies were installed with `npm install --ignore-scripts --no-audit --no-fund --package-lock=false --workspace @franken/live-bench` after `npm ci --ignore-scripts` failed because the existing lockfile is out of sync.

### DIFF
--- a/packages/live-bench/src/workspace/workspace-provisioner.ts
+++ b/packages/live-bench/src/workspace/workspace-provisioner.ts
@@ -1,3 +1,4 @@
+import { createHash } from 'node:crypto';
 import { cpSync, existsSync, lstatSync, mkdirSync, readdirSync, realpathSync, rmSync, writeFileSync } from 'node:fs';
 import { join, parse, relative, resolve, sep } from 'node:path';
 import { serializeToolCallEvidence } from '../evidence/tool-call-evidence.js';
@@ -162,12 +163,18 @@ function isLeapYear(year: number): boolean {
   return year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0);
 }
 
+const MAX_MODEL_PATH_SEGMENT_BYTES = 255;
+
 function modelPathSegment(model: string): string {
   const codeUnits = Buffer.alloc(model.length * 2);
   for (let index = 0; index < model.length; index += 1) {
     codeUnits.writeUInt16BE(model.charCodeAt(index), index * 2);
   }
-  return `model-${codeUnits.toString('hex')}`;
+  const encoded = `model-${codeUnits.toString('hex')}`;
+  if (Buffer.byteLength(encoded) <= MAX_MODEL_PATH_SEGMENT_BYTES) {
+    return encoded;
+  }
+  return `model-sha256-${createHash('sha256').update(codeUnits).digest('hex')}`;
 }
 
 function assertSafePathSegment(value: string, label: string): void {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2191, opened 2026-07-13
fix(web): surface dashboard response errors

### LINKED ISSUE #1695: fix(web): dashboard skill/security-profile errors show only "HTTP 400", discarding the server's actionable message — strict profile is an unrecoverable dead-end
**Finding**
The dashboard mutation client throws away the server's structured error body and surfaces only the HTTP status code, so the one flow that always fails without config (enabling the `strict` security profile) gives the operator no way to learn why it failed.

**Evidence**
- `packages/franken-web/src/lib/dashboard-api.ts:35`, `:45`, `:54` — `fetchSnapshot`, `toggleSkill`, and `updateSecurityProfile` all do `if (!res.ok) throw new Error(\`HTTP ${res.status}\`);` without reading the response body.
- `packages/franken-orchestrator/src/http/routes/security-routes.ts:60-66` — the server returns an actionable flat body: `{ error: 'Security profile "strict" requires allowedDomains to be configured' }` with 400 (same pattern at :41 and :77-80).
- `packages/franken-web/src/pages/dashboard-page.tsx:23-25`, `:303-309` — `describeError` returns `error.message`, so the alert reads `Could not save security profile strict; the profile was restored to the latest confirmed dashboard state. HTTP 400`, and `restoreSecurityProfile()` snaps the dropdown back.

Note the shape mismatch: these routes emit a flat `{ error: string }`, unlike the `{ error: { message } }` envelope the chat/network/beast clients already parse — so the fix needs both body-reading in `DashboardApiClient` and handling of the flat shape.

**Impact**
An operator selects **strict** in the SecurityPanel dropdown. The server rejects it because `allowedDomains` is not configured, the selector silently reverts, and the only feedback is "HTTP 400". The reason ("configure allowedDomains") never reaches the UI, so strict mode can never be enabled from the dashboard — a hard dead-end. The same message-stripping applies to skill toggle failures (`Could not enable <skill>… HTTP 404`).

**Suggested fix**
In `DashboardApiClient`, on `!res.ok` attempt `await res.json()` and prefer `body.error` (string) or `body.error.message` before falling back to `HTTP ${status}`; keep `describeError` as-is. Add a test asserting the strict-profile 400 message is shown verbatim in the alert.

---
**Other candidates (not filed this run)** — real but held back per noise/dedupe rules:
- `packages/franken-orchestrator/src/http/chat-server.ts:410-412` — `setSecurityConfig` persists the network config file with a truncating non-atomic `writeFileSync` while `atomicWriteFileSync` (`src/session/atomic-file.ts`) exists for exactly this; held back as theme-adjacent to closed #967 (atomic network state writes).
- `packages/franken-web/src/components/chat-shell.tsx:397-402` — initial `getConfig()` failure is swallowed (`.catch(() => undefined)`) and the Network config editor renders hardcoded defaults as if live, with no error and no re-fetch on Refresh (handler at :1257-1303 only re-fetches status/logs); held back as theme-adjacent to open #1018.
- `packages/franken-orchestrator/src/cli/beast-prompts.ts:10-17` — `beasts spawn` interview ignores each prompt's declared `required`/`options`, never displays valid choices, and only fails at the terminal `configSchema.parse()` with a raw Zod dump (same late-validation class as the init-wizard issue filed this run).
- `packages/franken-mcp-suite/src/cli/main.ts:231-290` — `fbeast mcp beast`/`mcp uninstall` are not wrapped in the error handler `mcp init` has (:219-229), so declining the beast-mode risk prompt (`beast-mode.ts:45`) surfaces as a raw unhandled-rejection stack trace.
- `packages/live-bench/src/cli/main.ts:24-36` + `src/corpus/loader.ts:55-57` — `fbeast-live-bench list <bad-path>` dumps a raw `ENOENT` `readdirSync` stack instead of "corpus root not found", and an existing-but-empty corpus dir prints nothing and exits 0.

djm204:
## Summary
- Read dashboard mutation/fetch error response bodies before throwing
- Prefer flat server `{ error: string }` and enveloped `{ error: { message } }` messages over bare HTTP status codes
- Add regression coverage for skill toggle and strict security-profile failures

## Test Plan
- npm run test -- src/lib/dashboard-api.test.ts
- npm run test -- src/pages/dashboard-page.test.tsx src/lib/dashboard-api.test.ts
- npm run typecheck
- npm run lint
- npm run build

Closes #1695

### DIFF
--- a/packages/franken-web/src/lib/dashboard-api.test.ts
+++ b/packages/franken-web/src/lib/dashboard-api.test.ts
@@ -95,6 +95,17 @@ describe('DashboardApiClient', () => {
       const client = new DashboardApiClient(BASE_URL);
       await expect(client.toggleSkill('missing', true)).rejects.toThrow('HTTP 404');
     });
+
+    it('prefers flat server error messages over bare HTTP status codes', async () => {
+      globalThis.fetch = vi.fn().mockResolvedValue({
+        ok: false,
+        status: 404,
+        json: async () => ({ error: 'Skill "missing" was not found' }),
+      });
+
+      const client = new DashboardApiClient(BASE_URL);
+      await expect(client.toggleSkill('missing', true)).rejects.toThrow('Skill "missing" was not found');
+    });
   });
 
   describe('updateSecurityProfile', () => {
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #2243, opened 2026-07-14
fix(orchestrator): remove unsafe double casts

### LINKED ISSUE #2226: fix(orchestrator): remove `as unknown as` double-casts on inbound comms and disabled-observer paths
## Finding
Two spots in franken-orchestrator use `as unknown as` double-casts that fully disconnect the compiler from the runtime shape — one on the untrusted inbound comms webhook path, one fabricating partial null-objects for the disabled-observer mode.

## Evidence
- `packages/franken-orchestrator/src/http/routes/comms-routes.ts:153` — the `/v1/comms/inbound` handler does:
  ```ts
  const body = validateBody(GenericCommsInboundBody, await parseJsonBody(c)) as unknown as ChannelInboundMessage;
  await gateway.handleInbound(body);
  ```
  `GenericCommsInboundBody` (comms-routes.ts:37–46) and `ChannelInboundMessage` (`src/comms/core/types.ts:5–14`) happen to match today, but nothing keeps them in sync — the double-cast silences the compiler entirely.
- `packages/franken-orchestrator/src/adapters/cli-observer-bridge.ts:128` and `:140` — the disabled-observer deps fabricate partial object literals forced through `as unknown as Trace` / `as unknown as Span`, omitting whatever other fields those interfaces require.

## Impact
If either the Zod schema or the `ChannelInboundMessage` interface drifts (field renamed, `receivedAt` type changed), the mismatch reaches `gateway.handleInbound()` as a runtime shape error on an externally-reachable webhook route instead of a compile error. Similarly, any consumer of the disabled-observer `Trace`/`Span` that reads a field beyond those provided gets `undefined` at runtime with no compiler warning — exactly the divergence the type system should catch if `Trace`/`Span` grow required fields.

## Suggested fix
- Declare the schema so the compiler checks the coupling (e.g. `const GenericCommsInboundBody = z.object({...}) satisfies z.ZodType<ChannelInboundMessage>` or derive the handler type via `z.infer`) and drop the cast at comms-routes.ts:153.
- Replace the partial literals in cli-observer-bridge.ts with a type-checked null-object factory for `Trace`/`Span` (explicit defaults for every required field) instead of double-casting.

## Other candidates (not filed this run)
- `packages/franken-observer/src/cost/CircuitBreaker.ts:40–47` swallows exceptions from trip handlers with zero diagnostics. The non-blocking contract is documented and deliberate, but a throwing HITL/alert handler currently fails with no trace at all — routing the caught error to a logger while continuing the loop would be a cheap improvement. Not filed because the behavior is documented-intentional; noted here for awareness.

djm204:
## Summary
- Type-couple generic comms inbound validation to `ChannelInboundMessage` with `satisfies z.ZodType<...>` and remove the unsafe `as unknown as ChannelInboundMessage` cast.
- Replace disabled CLI observer bridge trace/span double-casts with structurally complete typed null-object helpers.
- Add regression tests that guard against reintroducing the specific unsafe double-cast patterns.

## Test Plan
- `npm run test --workspace @franken/orchestrator -- tests/unit/http/comms-routes.test.ts tests/unit/adapters/cli-observer-bridge.test.ts`
- `npm run build --workspace @franken/types`
- `npm run build --workspace @franken/planner`
- `npm run build --workspace @franken/brain`
- `npm run build --workspace @franken/critique`
- `npm run build --workspace @franken/governor`
- `npm run build --workspace @franken/orchestrator`
- `npm run lint --workspace @franken/orchestrator` (0 errors; existing warnings only)
- `npm run test --workspace @franken/orchestrator`

Closes #2226

### DIFF
--- a/packages/franken-orchestrator/src/adapters/cli-observer-bridge.ts
+++ b/packages/franken-orchestrator/src/adapters/cli-observer-bridge.ts
@@ -27,6 +27,29 @@ interface ReplayCaptureRecord {
   readonly content: string;
 }
 
+function createDisabledTrace(traceId: string): Trace {
+  return {
+    id: traceId,
+    goal: 'tracing-disabled',
+    status: 'completed',
+    startedAt: Date.now(),
+    spans: [],
+  };
+}
+
+function createDisabledSpan(traceId: string, opts: { name: string; parentSpanId?: string }): Span {
+  return {
+    id: `tracing-disabled:${opts.name}`,
+    traceId,
+    ...(opts.parentSpanId ? { parentSpanId: opts.parentSpanId } : {}),
+    name: opts.name,
+    status: 'completed',
+    startedAt: Date.now(),
+    metadata: {},
+    thoughtBlocks: [],
+  };
+}
+
 export class CliObserverBridge implements IObserverModule {
   private readonly counter: TokenCounter;
   private readonly costCalc: CostCalculator;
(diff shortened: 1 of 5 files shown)


djm204:
@codex review

djm204:
@codex review

PR #2248, opened 2026-07-15
fix(orchestrator): write skill credentials atomically

### LINKED ISSUE #1697: fix(orchestrator): SkillCredentialStore rewrites .fbeast/.env with a non-atomic writeFileSync — a crash mid-write loses every stored credential
**Finding**
`SkillCredentialStore.setMany`/`remove` do a read-merge-then-truncating-rewrite of the entire credential file with a bare `writeFileSync` (no temp+rename, no fsync), even though the package already ships `atomicWriteFileSync` for exactly this pattern.

**Evidence**
- `packages/franken-orchestrator/src/skills/skill-credential-store.ts:60-66`:
  ```ts
  private writeEnv(entries: Record<string, string>): void {
    mkdirSync(dirname(this.envPath), { recursive: true });
    const lines = Object.entries(entries).map(
      ([key, value]) => `${key}=${value}`,
    );
    writeFileSync(this.envPath, lines.join('\n') + '\n');
  }
  ```
- `setMany` (:22-26) and `remove` (:29-33) both funnel the *full merged map* through this truncating rewrite.
- `readAll` → `parseEnv` (:40-58) silently skips malformed lines (`if (eqIndex === -1) continue;`), so a torn file doesn't error — it just quietly returns a partial/empty credential set, and the next `setMany` persists that loss permanently.
- The established helper exists in the same package: `packages/franken-orchestrator/src/session/atomic-file.ts` (`atomicWriteFileSync`, used by `network/network-state-store.ts` and the session stores).

**Impact**
A skill stores one new API key via `setMany`; the merged map contains all previously stored provider credentials; the process is killed (or disk fills) after `open(O_TRUNC)` but before the buffer is flushed. The `.fbeast/.env` file is left truncated, `readAll()` silently drops the mangled tail, and every credential after the tear point is gone — the user must re-enter all provider keys, with no error ever surfaced.

**Suggested fix**
Use `atomicWriteFileSync` from `src/session/atomic-file.ts` in `writeEnv`. While touching it, pass `{ mode: 0o600 }` for the credentials file (it currently inherits the default umask). Add a test that interrupted-write simulation (write temp, don't rename) leaves the previous file intact.

djm204:
## Summary
- Replace SkillCredentialStore's truncating .fbeast/.env rewrite with atomic temp-file + fsync + rename persistence.
- Extend atomicWriteFileSync with an optional mode and write credential env files as 0600.
- Add regression coverage proving SkillCredentialStore no longer depends on writeFileSync and creates owner-only env files.

## Test Plan
- npm test -- --run tests/unit/skills/skill-credential-store.test.ts tests/unit/session/atomic-file.test.ts
- npm run build --workspace @franken/types && npm run build --workspace @franken/observer && npm run build --workspace @franken/brain && npm run build --workspace @franken/critique && npm run build --workspace @franken/governor && npm run build --workspace @franken/planner
- npm run typecheck --workspace @franken/orchestrator
- npm run build --workspace @franken/orchestrator
- npm run lint --workspace @franken/orchestrator

Closes #1697

### DIFF
--- a/packages/franken-orchestrator/src/session/atomic-file.ts
+++ b/packages/franken-orchestrator/src/session/atomic-file.ts
@@ -39,13 +39,17 @@ function fsyncDir(dirPath: string): void {
  * torn/partial file, mirroring the pattern used by FileCheckpointStore.
  * The parent directory must already exist.
  */
-export function atomicWriteFileSync(filePath: string, contents: string): void {
+export function atomicWriteFileSync(
+  filePath: string,
+  contents: string,
+  options: { mode?: number } = {},
+): void {
   let tmpPath = `${filePath}.tmp.${writeCounter++}.${deterministicUuid('atomic-file-write')}`;
   try {
     let fd: number;
     for (;;) {
       try {
-        fd = openSync(tmpPath, 'wx');
+        fd = openSync(tmpPath, 'wx', options.mode);
         break;
       } catch (error) {
         if ((error as NodeJS.ErrnoException).code !== 'EEXIST') {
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #2322, opened 2026-07-16
fix(orchestrator): validate cached llm entries on read

### LINKED ISSUE #2312: fix(orchestrator): LlmCacheStore stamps schemaVersion on every write but never checks it on read, and serves unvalidated cache entries as LLM completions
**Finding**
`LlmCacheStore.read()` returns `JSON.parse(raw) as StoredCacheEntry` with no shape validation and no `schemaVersion` comparison, even though every save stamps `schemaVersion` — the field is written but dead. Its sibling store in the same directory does the check, so this is an inconsistency, not a policy.

**Evidence**
- `packages/franken-orchestrator/src/cache/llm-cache-store.ts:16,28` — every save writes `schemaVersion: this.options.schemaVersion`.
- `packages/franken-orchestrator/src/cache/llm-cache-store.ts:62-72` — `read()` does `return JSON.parse(raw) as StoredCacheEntry;` with no validation of shape or version.
- Contrast: `packages/franken-orchestrator/src/cache/provider-session-store.ts:33-35` — `if (stored.schemaVersion !== this.options.schemaVersion) return undefined;`.
- Consumer: `packages/franken-orchestrator/src/cache/cached-llm-client.ts:57-59` — `if (cachedResponse) { ...; return cachedResponse.content; }` — the value is returned directly as the LLM completion, typed `string`.

**Impact**
A cache entry written by an older schema, a hand-edited file, or valid-JSON-but-wrong-shape content (`{}`) is served as a hit: `content` is `undefined` typed as `string` and flows into planning/skill execution, where the first `.trim()`/`.length` throws a `TypeError` — or a stale/empty response is silently used and re-persisted. Bumping `CacheStoreOptions.schemaVersion` in a future release will not invalidate old entries, which is the field's entire purpose.

Related but distinct tracked work: #1219 (closed) added parse-failure tolerance and explicitly deferred schema validation ("If schema validation is added..."); #2031 (open) covers atomic *writes* for these stores. Neither covers the read-side version/shape check.

**Suggested fix**
In `LlmCacheStore.read()`, treat an entry as a miss (`return undefined`) when `entry.schemaVersion !== this.options.schemaVersion` or when `typeof entry.content !== 'string'`, mirroring `ProviderSessionStore.load()`. Add tests for an old-version entry and a wrong-shape entry.

djm204:
## Bug Description
- `LlmCacheStore` writes a `schemaVersion` with cached entries but reads the JSON payload without validating that it matches the current schema or that `content` is the expected string type.
- This means stale or malformed cache files could be treated as valid, causing runtime parse errors to surface only after cache reads.

## Root Cause
- `llm-cache-store.ts` used a direct `JSON.parse` cast to `StoredCacheEntry` and returned the object without checking the payload shape.
- The `schemaVersion` write path had no corresponding guard on read.

## Fix
- Added runtime validation in `LlmCacheStore.loadProjectEntry` and `loadSessionEntry`:
  - invalid JSON returns `undefined`.
  - wrong type shapes (missing `content`, non-string `content`, missing/invalid `schemaVersion`) return `undefined`.
  - mismatched `schemaVersion` returns `undefined`.
- Added regression tests in `llm-cache-store.test.ts` for:
  - schema version mismatch returning `undefined`.
  - malformed cache payloads returning `undefined`.
- Added a shared-lesson note to `tasks/resolve-issues-shared-lessons.md` for cache schema validation.

## Validation
- Ran targeted tests: `npm run test --workspace @franken/orchestrator -- tests/unit/cache/llm-cache-store.test.ts`
- Ran `npm run typecheck --workspace @franken/orchestrator` (fails in the worktree due shared workspace dependency environment; request re-check in CI/after dependency build).

Closes #2312

### DIFF
--- a/packages/franken-orchestrator/src/cache/llm-cache-store.ts
+++ b/packages/franken-orchestrator/src/cache/llm-cache-store.ts
@@ -62,12 +62,31 @@ export class LlmCacheStore {
   private async read(filePath: string): Promise<StoredCacheEntry | undefined> {
     try {
       const raw = await readFile(filePath, 'utf8');
-      return JSON.parse(raw) as StoredCacheEntry;
+      const stored = JSON.parse(raw) as unknown;
+
+      if (!this.isValidStoredCacheEntry(stored)) {
+        return undefined;
+      }
+
+      return stored;
     } catch (error) {
-      if ((error as NodeJS.ErrnoException).code === 'ENOENT' || error instanceof SyntaxError) {
+      if ((error as { code?: string }).code === 'ENOENT' || error instanceof SyntaxError) {
         return undefined;
       }
       throw error;
     }
+  }
+
+  private isValidStoredCacheEntry(value: unknown): value is StoredCacheEntry {
+    if (!value || typeof value !== 'object') {
+      return false;
+    }
+
+    const entry = value as Partial<StoredCacheEntry>;
+    return (
+      typeof entry.schemaVersion === 'number' &&
+      entry.schemaVersion === this.options.schemaVersion &&
+      typeof entry.content === 'string'
+    );
   }
 }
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

djm204:
@codex review

djm204:
@codex review

djm204:
@codex review

PR #2323, opened 2026-07-16
fix: preserve Codex hooks backup on invalid JSON

### LINKED ISSUE #2315: fix(mcp-suite): `fbeast mcp init --client=codex` silently overwrites an unparseable .codex/hooks.json, destroying the user's non-fbeast hooks
**Finding**
`writeCodexHooks` swallows a JSON parse failure of an existing `.codex/hooks.json` and then unconditionally rewrites the file, so a hooks file that fails a strict `JSON.parse` (trailing comma, comment, partial write) is treated as empty and replaced with an fbeast-only config — no warning, no backup, exit 0. This is an outlier against the recovery pattern the same file uses for the Claude client.

**Evidence**
- `packages/franken-mcp-suite/src/cli/init.ts:384-387` — `let existing = {}; ... try { existing = JSON.parse(readFileSync(hooksPath, 'utf-8')); } catch { /* ignore */ }`
- `packages/franken-mcp-suite/src/cli/init.ts:411` — `writeFileSync(hooksPath, JSON.stringify(existing, null, 2) + '\n');` runs regardless of the parse outcome.
- Three in-package contrasts prove the intended policy: (a) the Claude path in the same file uses `readJsonObjectFileOrRecover` (`init.ts:95,105`), which backs the invalid file up to a timestamped `.bak` and warns (`src/cli/settings-json.ts:17-31`); (b) that reader also tolerates comments/trailing commas via `stripJsonCommentsAndTrailingCommas`, which the codex path's bare `JSON.parse` does not; (c) the mirror `uninstall.ts:172-190` puts its write *inside* the try, so on parse failure it leaves the file untouched. Both helpers are already imported at `init.ts:17`.

**Impact**
A user with hand-edited `.codex/hooks.json` (one trailing comma is enough) runs `fbeast mcp init --client=codex`; their pre-existing PreToolUse/PostToolUse hook entries are silently destroyed and the CLI reports success. They discover the loss only when their other hooks stop firing, with no backup to restore from. (Closed #1236/#1047 covered shell-quoting of the hook command paths in this function — different defect.)

**Suggested fix**
Use the existing helpers: `existing = readJsonObjectFileOrRecover(hooksPath, readFileSync(hooksPath, 'utf-8'))` (backup + warn + lenient parse) and `writeJsonFileAtomic(hooksPath, existing)` instead of the bare `JSON.parse`/`writeFileSync`. Add a test that inits over a hooks.json with a trailing comma and asserts the original is preserved as a `.bak` and non-fbeast entries survive.

djm204:
## Summary
- Update Codex hook initialization to use recoverable JSON parsing when .codex/hooks.json is malformed.
- Persist hooks atomically using writeJsonFileAtomic for safer writes.
- Add coverage for invalid hooks file recovery and ensure a backup file is produced.

## Test Plan
- npm install (package dependencies)
- npm test src/cli/init.test.ts

Closes #2315

### DIFF
--- a/packages/franken-mcp-suite/src/cli/init.test.ts
+++ b/packages/franken-mcp-suite/src/cli/init.test.ts
@@ -22,6 +22,10 @@ function jsonBackups(dir: string): string[] {
   return readdirSync(dir).filter((name) => name.startsWith('settings.json.invalid-'));
 }
 
+function hooksBackups(dir: string): string[] {
+  return readdirSync(dir).filter((name) => name.startsWith('hooks.json.invalid-'));
+}
+
 describe('fbeast init', () => {
   const dirs: string[] = [];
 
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #2557, opened 2026-07-17
fix(memory): document memory query limit bounds

### LINKED ISSUE #2127: fix(memory): validate fbeast_memory_query limit input
## Description
`fbeast_memory_query` accepts `limit` as a string, converts it with `Number(args['limit'])`, and forwards the result directly into `brain.query`. Non-numeric input such as `limit: "banana"` becomes `NaN`; negative or fractional values are also accepted without normalization. The adapter then uses that value in both `brain.episodic.recall(input.query, input.limit ?? 20)` and `results.slice(0, input.limit ?? 20)`, so an invalid limit can silently produce no results or inconsistent truncation instead of a validation error/default.

## Affected locations
- `packages/franken-mcp-suite/src/servers/memory.ts:24` declares `limit` as a string input.
- `packages/franken-mcp-suite/src/servers/memory.ts:31-32` converts with `Number(...)` and forwards the value without checking `Number.isFinite`, integer-ness, or positive bounds.
- `packages/franken-mcp-suite/src/adapters/brain-adapter.ts:57` and `packages/franken-mcp-suite/src/adapters/brain-adapter.ts:80` use `input.limit ?? 20`; `NaN` is not nullish, so it is preserved into recall/slice.
- `packages/franken-mcp-suite/src/servers/memory.test.ts:45-47` only covers a valid numeric limit and does not assert invalid-limit behavior.

1. Call the memory MCP tool with a valid query that should match stored working memory, but pass `limit: "not-a-number"`.
2. `memory.ts` computes `Number("not-a-number")`, which is `NaN`.
3. `brain-adapter.ts` receives `{ limit: NaN }`; `input.limit ?? 20` stays `NaN` rather than defaulting to 20.
4. `results.slice(0, NaN)` coerces `NaN` to `0`, causing an empty result even when matching memory entries exist. Other odd values such as negative numbers can truncate from the end.

## Impact
This is a low-risk reliability bug: operator/tool callers can get misleading "No memory entries found" results solely because a limit value was malformed, making memory lookup look empty instead of surfacing bad input or falling back safely.

## Recommendation
Parse `limit` with a helper that accepts only positive finite integers, clamps to a documented maximum if needed, and either defaults to 20 or returns a validation error for malformed input. Add regression coverage for non-numeric, negative, zero, fractional, and valid string/numeric limits.

## Acceptance criteria
- `fbeast_memory_query` rejects or defaults invalid/non-finite `limit` values deterministically.
- Negative, zero, and fractional limits cannot silently alter result slicing.
- Tests cover invalid `limit` input and prove matching rows are not hidden by `NaN` conversion.
- The input schema documents the accepted limit shape.

Duplicate check: local open-issues cache terms `memory limit`, `fbeast_memory_query`, `memory query`, `NaN`, `Number(args`, and `limit` reviewed; related open issues #2065/#2042 cover rate-limiter numeric options, not memory-query result limits. Targeted all-state GitHub search `fbeast_memory_query limit Number(args NaN` returned no matches.

djm204:
## Summary
- Documents `fbeast_memory_query` limit bounds in the public MCP tool schema.
- Adds regression coverage so the schema stays aligned with the existing positive-integer validation.

## Test Plan
- npm ci
- npm test -- --run src/servers/memory.test.ts -t "limits memory type enums" (observed failing before the schema description update)
- npm test -- --run src/servers/memory.test.ts
- npm run build
- cd packages/franken-mcp-suite && npm run typecheck
- cd packages/franken-mcp-suite && npm run lint
- cd packages/franken-mcp-suite && npm run build

Closes #2127

### DIFF
--- a/packages/franken-mcp-suite/src/servers/memory.test.ts
+++ b/packages/franken-mcp-suite/src/servers/memory.test.ts
@@ -106,6 +106,10 @@ describe("Memory Server", () => {
       enum: ["working", "episodic"],
       description: "Filter by type: working or episodic",
     });
+    expect(queryTool.inputSchema.properties?.limit).toMatchObject({
+      type: "string",
+      description: "Max results as a positive integer from 1 to 1000 (default 20)",
+    });
   });
 
   it("delegates memory store/query/frontload/forget to the brain adapter", async () => {
--- a/packages/franken-mcp-suite/src/shared/tool-registry.ts
+++ b/packages/franken-mcp-suite/src/shared/tool-registry.ts
@@ -275,7 +275,7 @@ const TOOLS: ToolFull[] = [
       properties: {
         query: { type: 'string', description: 'Search query (substring match on key and value)' },
         type: { type: 'string', description: 'Filter by type: working or episodic', enum: ['working', 'episodic'] },
-        limit: { type: 'string', description: 'Max results (default 20)' },
+        limit: { type: 'string', description: 'Max results as a positive integer from 1 to 1000 (default 20)' },
         readScope: { type: 'string', description: 'Read scope: all (legacy), shared (hide agent-scoped entries), or agent (shared plus entries for agentId)', enum: ['all', 'shared', 'agent'] },
         agentId: { type: 'string', description: 'Agent id required when readScope is agent' },
       },


djm204:
@codex review

PR #2605, opened 2026-07-17
test(observer): cover invalid CircuitBreaker budgets

### LINKED ISSUE #1965: fix(observer): validate CircuitBreaker budget inputs
## Bug Description
`CircuitBreaker` accepts arbitrary numeric `limitUsd` and `spendUsd` values without validating that they are finite, non-negative amounts. The current implementation stores `options.limitUsd` directly and compares `spendUsd > this.limitUsd`, so values such as `NaN`, `Infinity`, or negative limits can produce misleading non-tripped states, permanently tripped states, or HITL payloads containing invalid spend/limit data.

## Affected locations
- `packages/franken-observer/src/cost/CircuitBreaker.ts:22-30` stores `limitUsd` and builds the result from raw numbers.
- `packages/franken-observer/src/cost/CircuitBreaker.ts:32-44` uses the raw comparison to trip/re-arm handlers.
- `packages/franken-observer/src/cost/CircuitBreaker.test.ts:11-31` covers normal below/equal/above values but has no invalid-number coverage.

## Reasoning path
1. Construct `new CircuitBreaker({ limitUsd: Number.NaN })` or call `breaker.check(Number.NaN)`.
2. JavaScript comparisons with `NaN` are always false, so the result reports `tripped: false` even though the budget input is unusable.
3. Construct `new CircuitBreaker({ limitUsd: -1 })`; ordinary non-negative spend immediately trips the breaker and can emit misleading limit metadata.
4. Use `Infinity` for either side and the guard can never trip or never recover depending on which value is infinite.

## Impact
Budget/HITL guardrails can silently accept corrupted telemetry and emit malformed alert payloads. This makes cost-safety signals harder to trust and can hide upstream cost-accounting bugs until operators debug odd breaker behavior manually.

## Recommendation
Validate `limitUsd` at construction and `spendUsd` in `check()` as finite, non-negative numbers. Add focused tests for `NaN`, `Infinity`, and negative values so invalid budget telemetry fails explicitly before tripping/re-arming state.

## Acceptance Criteria
- [ ] `CircuitBreaker` rejects non-finite or negative `limitUsd` during construction with a clear error.
- [ ] `CircuitBreaker.check()` rejects non-finite or negative `spendUsd` without mutating trip/re-arm state.
- [ ] Regression tests cover `NaN`, `Infinity`, and negative inputs for both constructor and `check()` paths.
- [ ] Existing valid limit behavior and rising-edge alert semantics remain unchanged.

Duplicate check: local issue cache plus targeted `gh issue list` searches for `CircuitBreaker limitUsd spendUsd NaN Infinity negative` returned no existing open issue covering this invalid-number invariant gap.

djm204:
## Summary
- Adds regression coverage for `CircuitBreaker` rejecting `-Infinity` budget values.
- Locks down that invalid `spendUsd` throws before emitting handlers or re-arming a tripped breaker.

## Test Plan
- [x] `npm run build --workspace @franken/types`
- [x] `npm run test --workspace @franken/observer -- CircuitBreaker.test.ts`
- [x] `npm run typecheck --workspace @franken/observer`
- [x] `npm run build --workspace @franken/observer`
- [x] `npm run lint --workspace @franken/observer` (passes with pre-existing warnings outside this change)
- [x] `git diff --check`

Closes #1965

### DIFF
--- a/packages/franken-observer/src/cost/CircuitBreaker.test.ts
+++ b/packages/franken-observer/src/cost/CircuitBreaker.test.ts
@@ -33,10 +33,25 @@ describe('CircuitBreaker', () => {
     it.each([
       ['NaN', Number.NaN],
       ['Infinity', Number.POSITIVE_INFINITY],
+      ['-Infinity', Number.NEGATIVE_INFINITY],
       ['negative values', -0.01],
     ])('throws RangeError for invalid spendUsd: %s', (_label, spendUsd) => {
       expect(() => breaker.check(spendUsd)).toThrow(RangeError)
       expect(() => breaker.check(spendUsd)).toThrow('spendUsd must be a finite non-negative number')
+    })
+
+    it('does not mutate trip state or emit handlers for invalid spendUsd', () => {
+      const handler = vi.fn()
+      breaker.on('limit-reached', handler)
+
+      breaker.check(0.51)
+      expect(handler).toHaveBeenCalledOnce()
+
+      expect(() => breaker.check(Number.NaN)).toThrow(RangeError)
+      expect(handler).toHaveBeenCalledOnce()
+
+      breaker.check(0.60)
+      expect(handler).toHaveBeenCalledOnce()
     })
   })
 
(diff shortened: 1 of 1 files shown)


djm204:
@codex review

PR #2645, opened 2026-07-18
fix(web): reset force restart state per tracked agent

### LINKED ISSUE #2643: Regression: tracked-agent Force Restart state leaks between agents
## Description
The Beasts tracked-agent detail panel keeps the Force Restart toggle in local component state, so once an operator enables Force for one running agent, the same detail panel instance can keep rendering Restart as the destructive force-restart confirmation when a different running agent is selected or the panel is closed/reopened without unmounting. That makes a dangerous action state leak across agents instead of being scoped to the current agent.

## Affected locations
- `packages/franken-web/src/components/beasts/agent-action-bar.tsx:89-90` initializes `forceRestart` with `useState(false)` but has no reset keyed to the agent id, panel open state, or status.
- `packages/franken-web/src/components/beasts/agent-action-bar.tsx:104-149` switches the Restart button into the force-confirmation flow while `forceRestart` remains true.
- `packages/franken-web/src/components/beasts/agent-detail-panel.tsx:136-147` passes only status/handlers to `AgentActionBar`, so the action bar cannot distinguish one selected agent from another and reset its local dangerous-mode state.
- `packages/franken-web/src/components/beasts/agent-detail-panel.test.tsx:59-120` covers edit-store reset behavior, but there is no regression coverage for action-bar Force state when the selected agent changes.

## Impact
An operator can enable Force Restart while inspecting one running agent, then switch to another running agent and see the Restart control already in force mode. The confirmation dialog still appears, but the UI has silently carried a destructive-mode preference to a different target, increasing the chance of interrupting the wrong agent mid-turn.

## Recommendation
Pass a stable `agentId` or reset key into `AgentActionBar` and clear `forceRestart` whenever the selected agent, panel open state, or non-running status changes. Add a React Testing Library regression that toggles Force for one running detail, rerenders the panel with a different running agent, and asserts Restart returns to the normal non-force state.

## Duplicate check
Checked the open issue cache and GitHub searches for `force restart`, `AgentActionBar`, `tracked agent`, and `dashboard restart`. No open duplicate was found; all-state search only surfaced closed confirmation-hardening issues #1140/#1159.

djm204:
## Summary
- resets the tracked-agent action bar Force Restart state when the selected agent, panel-open reset key, or status changes
- passes the detail panel's selected agent/open state into the action bar reset boundary
- adds a React Testing Library regression for switching from one running agent to another after enabling Force

## Test Plan
- npm test -- tests/components/beasts/agent-detail-panel.test.tsx
- npm test -- tests/components/beasts/agent-action-bar.test.tsx tests/components/beasts/agent-detail-panel.test.tsx
- npm run lint --workspace @franken/web
- npm run typecheck --workspace @franken/web
- npm run build --workspace @franken/web

Closes #2643

### DIFF
--- a/packages/franken-web/src/components/beasts/agent-action-bar.tsx
+++ b/packages/franken-web/src/components/beasts/agent-action-bar.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import * as AlertDialog from '@radix-ui/react-alert-dialog';
 
 export type AgentLifecycleAction = 'start' | 'stop' | 'restart' | 'resume' | 'delete' | 'kill';
@@ -7,6 +7,7 @@ interface AgentActionBarProps {
   status: string;
   hasLinkedRun: boolean;
   agentLabel?: string;
+  resetKey?: string;
   pendingAction?: AgentLifecycleAction | null;
   onStart: () => void;
   onStop: () => void;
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #3230, opened 2026-07-18
test(sandbox): require valid Dockerfile user directive

### LINKED ISSUE #3228: Testing: Fail sandbox Dockerfile check when USER directive is missing
## Description
The sandbox Dockerfile test is intended to enforce non-root container execution, but it uses optional chaining when extracting the parsed `USER` value. If the Dockerfile omits a `USER` directive entirely, `userLine` becomes `undefined`, and the assertion `expect(undefined).not.toBe('0')` passes. This allows the test to succeed without proving the Dockerfile enforces a non-root user.

## Affected locations
- `tests/sandbox-dockerfile.test.ts:50-52`

## Impact
A Dockerfile missing a `USER` directive (or malformed user syntax) can pass the current test suite while still running as root in sandboxed execution, reducing container hardening guarantees and masking a meaningful regression in security posture.

## Recommendation
Update the test to fail explicitly when `USER` is missing or unparsable, then validate the parsed user ID is not `0`. For example, assert `userLine` exists before split, and separately validate the numeric/username form used by the Dockerfile parser.

djm204:
## Summary
- require the sandbox Dockerfile to contain a syntactically valid `USER` directive
- validate the parsed user principal rather than allowing missing or malformed values through optional chaining
- preserve explicit rejection of root by name or UID 0

## Test plan
- `npx vitest run tests/sandbox-dockerfile.test.ts -t "declares a non-root default container UID"`
- mutant checks: removed `USER` and malformed `USER :10001` both fail the targeted assertion
- `npm run test:root` (573 passed, 1 skipped)
- `npm run typecheck`
- `npm run lint` (passes with pre-existing warnings)
- `npm run build`

Closes #3228

### DIFF
--- a/tests/sandbox-dockerfile.test.ts
+++ b/tests/sandbox-dockerfile.test.ts
@@ -120,11 +120,10 @@ describe('sandbox Dockerfile', () => {
   }, 60_000);
 
   it('declares a non-root default container UID', () => {
-    const userLine = dockerfile.split('\n').find((line) => line.startsWith('USER '));
+    const userDirective = dockerfile.match(/^USER\s+([^\s:]+)(?::[^\s]+)?\s*$/mu);
 
-    expect(userLine).toBeDefined();
-    expect(userLine).not.toBe('USER root');
-    expect(userLine).not.toBe('USER 0');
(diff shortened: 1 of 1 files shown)


djm204:
@codex review

PR #3244, opened 2026-07-18
perf(mcp): validate only audit trail tail on append

djm204:
## Summary
- replace full audit-chain verification on every append with indexed tail validation
- validate and insert inside one immediate SQLite transaction
- preserve legacy-row compatibility and explicit full-chain migration/verification
- cover long trails, damaged tails, and legacy tails

## Test plan
- npm test -- --run src/adapters/observer-adapter.test.ts
- npm test
- npm run typecheck
- npm run lint
- npm run build
- root npm run build
- git diff --check

Closes #3202

Issue #3202, opened 2026-07-18
Performance: Avoid full audit verification on every append


djm204:
## Description
Every MCP audit append preflights by verifying the existing session trail before inserting the next row.

## Affected locations
- `packages/franken-mcp-suite/src/adapters/observer-adapter.ts:75-84` calls `verifyAuditTrail(... migrate: true ...)` inside `log()`.
- `packages/franken-mcp-suite/src/adapters/observer-adapter.ts:85-106` then reads the last hash and inserts one row.

## Impact
Audit logging becomes O(n²) over a long session and can make high-volume tool use progressively slower.

## Recommendation
Replace per-append full verification with constant-time last-row validation plus a separate explicit verify/repair operation.

## Acceptance criteria
- Appending an audit event is O(1) with respect to trail length.
- Explicit full verification still exists for operators.
- Bench/unit tests cover long sessions.

## LLM handoff notes
Use the cited source paths as the owning implementation seams; keep the change implementation-sized and add focused unit coverage for the failing behavior.

PR #3245, opened 2026-07-18
fix(mcp): preserve audit integrity during migration

djm204:
## Summary
- hold an immediate SQLite transaction from legacy-chain inspection through migration
- reject legacy-16 downgrade tails after current full-hash parents
- add deterministic competing-writer and downgrade regression tests

## Context
Follow-up to #3244 for two late independent review findings on issue #3202.

## Test plan
- npm test -- --run src/adapters/observer-adapter.test.ts (25 passed)
- npm test (546 passed)
- npm run typecheck
- npm run lint
- npm run build
- git diff --check


djm204:
@codex review

PR #3255, opened 2026-07-19
fix(web): surface initial network config load failures

### LINKED ISSUE #3195: fix(web): initial network-config load failure is silently swallowed — editor renders fabricated defaults
**Finding**
The initial `getConfig()` fetch in `ChatShell` swallows failures with an empty catch, so when the config endpoint fails the Network page config editor silently renders the hard-coded client-side placeholder values as if they were the server's real configuration.

**Evidence**
`packages/franken-web/src/components/chat-shell.tsx:226-233`:

```ts
const configRequestId = ++networkConfigRequestIdRef.current;
void client.getConfig()
  .then((nextConfig) => {
    if (configRequestId === networkConfigRequestIdRef.current) {
      setNetworkConfig(nextConfig);
    }
  })
  .catch(() => undefined);
```

On failure, `networkConfig` stays at the placeholder initialized at `chat-shell.tsx:135-139` (`{ network: { mode: 'secure', secureBackend: 'local-encrypted' }, chat: { model: 'claude-sonnet-4-6', enabled: true, host: '127.0.0.1', port: 3737 } }`) and `networkConfigError` is never set. The manual-refresh path in the same file (`chat-shell.tsx:~1135-1146`) handles the identical failure correctly with `setNetworkConfigError('Unable to refresh network config: ...')`, so the initial-load gap is an oversight, not a policy.

**Impact**
When only the config endpoint fails (the parallel `getStatus()` failure at line 220 does surface a banner, so a fully-unreachable server is covered), an operator sees a config editor pre-filled with plausible-but-fabricated values and no error. They can edit and save from that fabricated baseline, unknowingly overwriting real server config (e.g. resetting the chat model or port to the client-side defaults).

**Suggested fix**
Mirror the refresh path: add `.catch((error) => { if (configRequestId === networkConfigRequestIdRef.current) setNetworkConfigError(\`Unable to load network config: ${networkErrorMessage(error, 'Request failed.')}\`); })`, and consider representing "config not yet loaded" as null so the editor renders a load-error state instead of placeholder values.

Related but distinct open issues (checked before filing): #2620 covers the editor not resyncing/stale saves, #3152/#2795 cover missing *save* pending/failure feedback; none cover the swallowed *load* error.

djm204:
## Summary
- represent network config as unloaded until the API returns real server data
- surface initial config fetch failures and keep the config editor unavailable instead of showing fabricated defaults
- add regression coverage for the initial failure path

## Test plan
- `npm test` (packages/franken-web; 641 passed)
- `npm run typecheck` (packages/franken-web)
- `npm run lint` (packages/franken-web; 0 errors, 18 pre-existing warnings)
- `npm run build` (packages/franken-web)

Closes #3195

### DIFF
--- a/packages/franken-web/src/components/chat-shell.test.tsx
+++ b/packages/franken-web/src/components/chat-shell.test.tsx
@@ -307,6 +307,17 @@ describe('ChatShell route heading', () => {
     expect(screen.queryByDisplayValue('stale-model')).toBeNull();
   });
 
+  it('surfaces an initial config load failure without rendering fabricated defaults', async () => {
+    networkApiMocks.getStatus.mockResolvedValue({ mode: 'secure', secureBackend: 'local-encrypted', services: [] });
+    networkApiMocks.getConfig.mockRejectedValueOnce(new Error('HTTP 503'));
+
+    render(<ChatShell baseUrl="http://localhost:3737" projectId="default" version="0.2.1" />);
+
+    expect((await screen.findByRole('alert')).textContent).toContain('Unable to load network config: HTTP 503');
+    expect(screen.queryByRole('button', { name: 'Save config' })).toBeNull();
+    expect(screen.queryByDisplayValue('claude-sonnet-4-6')).toBeNull();
+  });
+
   it('surfaces failed config refreshes from the Network page', async () => {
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #3302, opened 2026-07-19
fix(compose): tolerate slow Chroma startup

### LINKED ISSUE #3281: Harden docker-compose startup checks against slow environment startup delays
## Description
The Compose definition uses a `timeout: 5s` value in healthcheck-related configuration (`docker-compose.yml:15-17`). A 5-second probe timeout can be too short for heavier services during first startup, especially on slower or resource-constrained CI/dev hosts, producing false-unhealthy results even when services are otherwise healthy.

## Affected locations
- `docker-compose.yml:15-17` (`timeout: 5s`)

## Impact
Intermittent startup instability can occur for the local/dev compose stack and any deployment path that relies on this file, increasing false failures, extra retries, and operator/operator automation churn.

## Recommendation
Increase the relevant service healthcheck timeout(s) from `5s` to a more realistic startup budget and tune per-service `start_period` and retry settings to match actual warm-up behavior. Add a short note in `docker-compose.yml` documenting expected startup characteristics so future changes keep these checks stable.

djm204:
## Summary
- give Chroma health probes a 15-second response timeout
- add a 30-second startup grace period and five retries for slower hosts
- document and regression-test the startup budget

## Validation
- `npm run test:root -- --run tests/local-setup-scripts.test.ts` (31 passed)
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- parsed `docker-compose.yml` with `js-yaml` and asserted the healthcheck values

`docker compose config --quiet` could not run because Docker is not installed on this worker.

Closes #3281

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,9 +18,11 @@ services:
           '-ec',
           'exec 3<>/dev/tcp/127.0.0.1/8000 && printf "GET /api/v2/heartbeat HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n" >&3 && grep -q "200 OK" <&3',
         ]
+      # ChromaDB can take longer to initialize on slow or resource-constrained hosts.
       interval: 10s
-      timeout: 5s
-      retries: 3
+      timeout: 15s
+      retries: 5
+      start_period: 30s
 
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #3383, opened 2026-07-19
fix(orchestrator): remove stale banner dependencies

### LINKED ISSUE #3361: chore(orchestrator): banner rewrite orphaned the `sharp` optional dependency and the documented FRANKENBEAST_PLAIN_BANNER flag
**Finding**
Commit `40f680d` ("feat(cli): simplify startup banner", 2026-07-19) replaced the graphic/image banner path with a static ASCII wordmark, but left behind (a) `sharp` as a shipped dependency with zero remaining importers, and (b) operator documentation for a `FRANKENBEAST_PLAIN_BANNER` environment variable that the CLI now silently ignores. The banner also no longer prints the CLI version, which the old fallback banner always did.

**Evidence**
- `packages/franken-orchestrator/src/logging/beast-logger.ts:85` — `renderBanner(_root)` now unconditionally returns the static `BANNER` constant; the `sharp` dynamic import, `shouldRenderGraphicBanner()`, `FRANKENBEAST_PLAIN_BANNER` check, and `readBannerVersion()` were all deleted in `40f680d`.
- `grep` for `from 'sharp'` / `import('sharp')` / `require('sharp')` across the repo: **0 matches**. The only remaining references are the dependency declarations themselves.
- `packages/franken-orchestrator/package.json:69` — `optionalDependencies: { "sharp": "^0.35.3" }` still ships with the published CLI (`files: ["dist"]` — optional deps are installed by consumers).
- `package.json:71` (root) — `sharp: ^0.35.3` still in root devDependencies.
- `README.md:555` — "Set `FRANKENBEAST_PLAIN_BANNER=1` to force the CLI startup banner to use the plain ASCII fallback instead of the image-rendered graphic banner…" describes behavior that no longer exists (there is no graphic banner and the variable is never read).
- `.env.example:86` — still lists `FRANKENBEAST_PLAIN_BANNER=0` under "Advanced local diagnostics".

**Impact**
Every `npm install` of the published CLI still downloads `sharp`, a heavy native module with per-platform binaries, for a code path that no longer exists — pure install bloat plus an unnecessary native-module supply-chain/postinstall surface (the exact class of problem `scripts/check-phantom-deps.mjs` was written for, in reverse). Operators following README.md:555 set an env var that now does nothing: CI logs and log processors that relied on the documented plain-banner escape hatch silently get the ANSI-colored Unicode banner instead. Startup output also no longer shows which CLI version is running.

**Suggested fix**
Remove `sharp` from `packages/franken-orchestrator/package.json` optionalDependencies and from root devDependencies (re-lock); delete the `FRANKENBEAST_PLAIN_BANNER` paragraph from README.md and the `.env.example` entry, or reinstate the flag as a "suppress ANSI banner" toggle if the escape hatch is still wanted; consider restoring the version line under the new wordmark.

djm204:
## Summary
- remove the obsolete root and orchestrator `sharp` declarations plus their lockfile artifacts
- remove stale `FRANKENBEAST_PLAIN_BANNER` documentation from the README and environment example
- add a focused regression test covering package metadata, lockfile declarations, and documentation

## Validation
- `npm exec -- vitest run tests/issue-3361-banner-dependency.test.ts` (2 passed)
- `npm run lint --workspace @franken/orchestrator` (0 errors; existing warnings only)
- `npx turbo run build --filter=@franken/orchestrator...` (7/7 tasks passed)
- `npm pack --dry-run --workspace @franken/orchestrator --json`
- `npm test --workspace @franken/orchestrator` (4,100/4,101 passed; one timing-sensitive planner deadline assertion failed, then passed in isolation)
- `npm exec -- vitest run tests/unit/llm-graph-builder.test.ts -t 'checks the deadline after synchronous stage parsing completes'` (passed)

Closes #3361

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,7 +83,6 @@ FRANKEN_MIN_CRITIQUE_SCORE=0.7
 # FRANKENBEAST_ALLOW_NONINTERACTIVE_APPROVAL=0
 
 # ── Advanced local diagnostics ──
-# FRANKENBEAST_PLAIN_BANNER=0
 # Internal marker set by `frankenbeast network` for managed child services.
 # Leave unset for normal standalone debugging. When set to literal "1", managed
 # children suppress the CLI startup banner and `chat-server` requires an operator
--- a/README.md
+++ b/README.md
@@ -544,8 +544,6 @@ If `stateDir` is set to a path inside a sibling Hermes profile such as `.hermes/
 
 ### Operator environment variables
 
-Set `FRANKENBEAST_PLAIN_BANNER=1` to force the CLI startup banner to use the plain ASCII fallback instead of the image-rendered graphic banner. This is useful for CI logs, terminals with limited image rendering support, and log processors that should receive a text-only banner layout. The fallback banner may still include ANSI color codes; leave the variable unset, or set it to any value other than `1`, to keep the normal graphic banner path.
-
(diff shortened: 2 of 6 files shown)


djm204:
@codex review

PR #3404, opened 2026-07-19
docs(readme): link package map before architecture

### LINKED ISSUE #3402: Docs: Point architecture link to package map before legacy MOD diagrams
## Description
The README warns that its opening Beast-loop diagrams still use legacy MOD labels and that the package inventory table is the authoritative source for current package mapping. Immediately after that warning, it sends readers to `docs/ARCHITECTURE.md` for the full interconnection diagram without first directing them to the package inventory/table. This can make the first architecture path reinforce the legacy MOD terminology the README just warned readers not to treat as authoritative.

## Affected locations
- `README.md:28-30` — warns that the diagrams still use MOD labels and says the package inventory table is authoritative.
- `README.md:30-32` — sends readers directly to `docs/ARCHITECTURE.md` for the full interconnection diagram.

## Impact
New contributors following the top-of-file architecture path can confuse legacy capability labels with current workspace/package names, especially before they reach the later package inventory. That increases onboarding friction and can lead to stale terminology in docs, issues, or implementation plans.

## Recommendation
Adjust the README introduction so the architecture link is paired with, or follows, an explicit pointer to the current package inventory/table. For example, tell readers to use the package inventory as the package-name source of truth before consulting the full architecture diagram, or add a short note near the architecture link that the diagram may use legacy capability labels while the inventory maps them to current packages.

djm204:
## Summary
- link the current workspace package inventory before the architecture diagram
- keep the package inventory explicit as the package-name source of truth

## Verification
- `git diff --check`
- verified the inventory link precedes the architecture link and targets the existing `#current-workspace-packages` anchor

Closes #3402

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Frankenbeast is currently organized as 10 npm workspace packages under `packages
 
 The diagrams below use current package names or implementation-surface names, and the package inventory table remains the authoritative workspace map.
 
-See [docs/ARCHITECTURE.md](docs/ARCHITECTURE.md) for the full interconnection diagram.
+Start with the [current workspace package inventory](#current-workspace-packages) for the package-name source of truth, then see [docs/ARCHITECTURE.md](docs/ARCHITECTURE.md) for the full interconnection diagram.
 
 ```mermaid
 flowchart TD


djm204:
@codex review

PR #3408, opened 2026-07-19
perf(observer): cache loop detector comparisons

### LINKED ISSUE #2679: LoopDetector repeats expensive span normalization and Levenshtein checks
## Description
`LoopDetector.spanNamesMatch()` normalizes span names during every comparison and falls back to Levenshtein distance for non-exact matches inside the sliding-window scan.

## Affected locations
`packages/franken-observer/src/incident/LoopDetector.ts:118-190`

## Impact
Long span names or large history windows can make loop detection CPU-heavy on every completed span, blocking normal trace processing.

## Recommendation
Normalize/digit-mask span names once on ingestion and cache derived values in the history buffer. Add tests or benchmarks proving comparison work stays bounded for large windows.

## Acceptance criteria
- The affected behavior is covered by a focused regression test or component test.
- The implementation handles the documented edge case without weakening existing contracts.
- LLM handoff: start at the listed files, reproduce from the cited code path, then keep the fix implementation-sized and package-local unless shared helpers are already present.

djm204:
## Summary
- normalize and digit-mask each span name once when it enters LoopDetector history
- cache pairwise span-name comparison results with weak references so repeated scans avoid duplicate Levenshtein work
- preserve original span names in detected patterns and add focused bounded-work regression coverage

## Test plan
- `npm test -- --run src/incident/LoopDetector.test.ts`
- `npm test --workspace @franken/observer`
- `npm run typecheck --workspace @franken/observer`
- `npm run lint --workspace @franken/observer` (passes with 6 pre-existing warnings)

Closes #2679

### DIFF
--- a/packages/franken-observer/src/incident/LoopDetector.test.ts
+++ b/packages/franken-observer/src/incident/LoopDetector.test.ts
@@ -188,6 +188,37 @@ describe('LoopDetector', () => {
     })
   })
 
+  describe('comparison caching', () => {
+    it('normalizes each span once and reuses pairwise fuzzy-match results across scans', () => {
+      const internals = LoopDetector as unknown as {
+        normalizeSpanName(spanName: string): string
+        levenshteinDistance(left: string, right: string): number
+      }
+      const normalizeSpy = vi.spyOn(internals, 'normalizeSpanName')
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #3421, opened 2026-07-20
test(root): validate aggregate task wiring

### LINKED ISSUE #3420: Testing: Make verify-everything validate task wiring instead of only checking line presence
## Description
`tests/verify-everything.test.ts` appears to validate the repository's verification task wiring with a presence-only assertion. Checking that `tasksLine` is defined can pass even when the underlying command is stale, incomplete, or points at the wrong task, so the test does not fully protect the verification entrypoint it is meant to guard.

## Affected locations
- `tests/verify-everything.test.ts:35-37` — `expect(tasksLine).toBeDefined();`

## Impact
A broken or incomplete verify-everything task can survive CI as long as the searched line exists. That weakens the safety net around the repo's aggregate verification path and can let regressions land without running the intended checks.

## Recommendation
Parse the relevant package script or task configuration and assert the expected command components explicitly. The test should fail when required subtasks are missing, renamed, or no longer wired into the verify-everything path, rather than only confirming that a matching line was found.

djm204:
## Summary
- parse the root package and Turbo configurations in verify-everything
- require build/test scripts to map to their matching Turbo tasks
- cover missing, stale, and invalid task wiring without recursively running Turbo

## Test Plan
- npm run test:root
- npm run typecheck
- npm run lint
- npm run build

Closes #3420

### DIFF
--- a/tests/helpers/verification-task-wiring.ts
+++ b/tests/helpers/verification-task-wiring.ts
@@ -1,0 +1,28 @@
+type ScriptMap = Record<string, unknown>;
+type TurboTaskMap = Record<string, unknown>;
+
+const REQUIRED_TASKS = ['build', 'test'] as const;
+
+export const validateVerificationTaskWiring = (
+  scripts: ScriptMap,
+  turboTasks: TurboTaskMap,
+): void => {
+  for (const task of REQUIRED_TASKS) {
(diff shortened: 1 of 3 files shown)


djm204:
@codex review

PR #3458, opened 2026-07-20
chore(types): enforce workspace strictness

### LINKED ISSUE #2313: chore(tsconfig): strictness drift — franken-observer, franken-types, and live-bench compile without the safety flags every other package enforces
**Finding**
There is no shared tsconfig base; each package hand-copies its compiler flags, and three packages have silently drifted to bare `"strict": true` while the rest of the monorepo enforces four additional safety flags. The root `tsconfig.test.json` that would re-check most packages under the strict flag set is referenced only by a structure test and is never actually compiled by any script or workflow.

**Evidence**
- `packages/franken-brain/tsconfig.json` (and critique, governor, mcp-suite, orchestrator) set `strict`, `noUncheckedIndexedAccess`, `exactOptionalPropertyTypes`, `noImplicitReturns`, `noFallthroughCasesInSwitch`.
- `packages/franken-observer/tsconfig.json`, `packages/franken-types/tsconfig.json`, `packages/live-bench/tsconfig.json` — only `"strict": true`; all four extra flags missing. Partial drift: `packages/franken-web/tsconfig.json` lacks `exactOptionalPropertyTypes`; `packages/franken-planner/tsconfig.json` lacks `noFallthroughCasesInSwitch`.
- Enforcement is per-package `tsc --noEmit` via `turbo run typecheck` (root `package.json:14`); root `tsconfig.json` has `include: []` (paths host only).
- `tsconfig.test.json` is referenced nowhere except `tests/tsconfig-paths.test.ts` (no `tsc -p tsconfig.test.json` in any package.json script, turbo.json task, or `.github/workflows/*`), and `live-bench`/`franken-mcp-suite` are allowlisted out of its includes anyway (`tests/tsconfig-paths.test.ts:64-68`).

**Impact**
`@franken/types` is the shared contract package: with `exactOptionalPropertyTypes` off it can build objects with explicitly-`undefined` optional properties that downstream packages (compiled with the flag on) assume cannot exist. In observer, indexed access (`arr[i]`) typechecks as non-undefined while every sibling package would be forced to guard. Latent bugs land preferentially in the weakest-checked packages and surface at package boundaries. (Closed #1071 covered deriving `tsconfig.test.json` *includes* from workspaces; the flag drift and the unenforced strict pass are not tracked anywhere.)

**Suggested fix**
Add a shared `tsconfig.base.json` with the full strict flag block and have every package extend it; add the missing flags to observer/types/live-bench (plus `exactOptionalPropertyTypes` to web and `noFallthroughCasesInSwitch` to planner), fixing the violations that surface. Then either wire `tsc -p tsconfig.test.json` into an enforced root script/CI step or delete the file so it stops implying coverage it doesn't provide.

djm204:
## Summary
- enforce the five required strict compiler options across every TypeScript workspace
- add regression coverage to prevent future tsconfig drift
- resolve strictness findings without weakening runtime contracts

## Verification
- `npm run test:root -- tests/tsconfig-paths.test.ts`
- package tests for `@franken/types`, `@franken/observer`, `@franken/live-bench`, and `@franken/web` (one existing web timing test passed on focused rerun)
- `npm run typecheck`
- `npm run build`
- `npm run lint`
- hardcoded-secret scan
- local Codex review clean after one finding was fixed

Closes #2313


djm204:
@codex review

PR #3566, opened 2026-07-22
fix(governor): validate Slack timestamp integers

### LINKED ISSUE #3537: Validation: Missing numeric validation for timestamp in Governor HTTP handler
## Description
**Quality goal:** Runtime validation at trust boundaries (`runtime-validation`)

The Governor HTTP server parses a `timestamp` field from incoming JSON payloads and converts it with `Number(timestamp)` without any validation. If the supplied value is non‑numeric, missing, or out of expected bounds, `tsSeconds` becomes `NaN` or an invalid value, which can propagate through the system and cause runtime errors or incorrect scheduling behavior.

## Affected locations
- `packages/franken-governor/src/server/app.ts` lines 267‑269 (conversion of `timestamp` to a number).

## Impact
- Unvalidated timestamps can lead to `NaN` values, causing downstream calculations to fail silently or produce incorrect results.
- Potential denial‑of‑service if malformed timestamps trigger unhandled exceptions.
- Reduces reliability and violates the Runtime validation quality goal which requires numeric bounds to be checked before use.

## Recommendation
- Introduce explicit runtime validation for the `timestamp` field, e.g., using Zod:
  ```ts
  const TimestampSchema = z.preprocess((arg) => Number(arg), z.number().int().nonnegative());
  const parsed = TimestampSchema.safeParse(requestBody.timestamp);
  if (!parsed.success) {
    return new Response(JSON.stringify({ error: 'Invalid timestamp' }), { status: 400 });
  }
  const tsSeconds = parsed.data;
  ```
- Alternatively, add manual checks: verify `!Number.isNaN(tsSeconds)` and that it falls within acceptable range (e.g., non‑negative and reasonable upper bound).
- Return a clear 400 Bad Request error with an explanatory message when validation fails.
- Add unit tests covering valid and invalid timestamp scenarios.

djm204:
## Summary
- require Slack request timestamps to parse as safe integers before replay-skew checks
- reject non-numeric and fractional timestamp headers with the existing authentication error
- add targeted regression coverage for both invalid forms

## Verification
- `npm run test --workspace @franken/governor` (336 passed)
- `npm run typecheck --workspace @franken/governor`
- `npm run lint --workspace @franken/governor` (0 errors; 3 pre-existing warnings)
- `npm run build --workspace @franken/governor`
- `npm run build`

Closes #3537

### DIFF
--- a/packages/franken-governor/src/server/app.ts
+++ b/packages/franken-governor/src/server/app.ts
@@ -545,7 +545,7 @@ export function createGovernorApp(options: GovernorAppOptions = {}): Hono {
     // Reject stale timestamps to mitigate replay attacks.
     const tsSeconds = Number(timestamp);
     if (
-      !Number.isFinite(tsSeconds) ||
+      !Number.isSafeInteger(tsSeconds) ||
       Math.abs(Date.now() / 1000 - tsSeconds) > SLACK_MAX_TIMESTAMP_SKEW_SECONDS
     ) {
       return c.json({ error: { message: 'Stale or invalid Slack timestamp' } }, 401);
--- a/packages/franken-governor/tests/unit/server/app.test.ts
+++ b/packages/franken-governor/tests/unit/server/app.test.ts
@@ -708,6 +708,24 @@ describe('Governor Hono Server', () => {
       expect(res.status).toBe(401);
     });
 
+    it.each([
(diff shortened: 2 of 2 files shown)


djm204:
@codex review

PR #3581, opened 2026-07-22
test(orchestrator): make process stop assertions deterministic

### LINKED ISSUE #3576: Flaky CI: ProcessBeastExecutor stop tests fail when a real PID 777 exists on the runner
## Symptom

`build-test-lint` intermittently fails on two tests in `packages/franken-orchestrator/tests/unit/beasts/process-beast-executor.test.ts`:

- `stops the current attempt without deleting the run row`
- `marks run as stopped when the process exits from an operator stop signal`

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 777, { processGroupOwned: true } ]
Received:
  1st vi.fn() call:
  [ 777, { "expectedStartTimeTicks": "1155", "processGroupOwned": true } ]
```

Observed today on `main` itself: run [29926917636](https://github.com/djm204/frankenbeast/actions/runs/29926917636) (commit `1f2b38c8`, whose diff touches only `franken-mcp-suite`) fails both tests, while runs `9a8f24f1`/`3c50cb18` minutes earlier pass. Also reproduced on PR #3462's run [29924325617](https://github.com/djm204/frankenbeast/actions/runs/29924325617) (diff touches only the agent-create body cap). Within a failing run all `CI_TEST_RETRIES` retries fail identically; across runs it comes and goes.

## Mechanism

Both tests stub the supervisor with `spawn: vi.fn(async () => ({ pid: 777 }))` — a hardcoded fake PID — and then assert an **exact** options object:

```ts
expect(supervisor.stop).toHaveBeenCalledWith(777, { processGroupOwned: attempt.executorMetadata?.processGroupOwned === true });
```
(test file lines ~909 and ~964)

But `ProcessBeastExecutor.start()` derives real metadata from the host for that fake PID: `processStartTimeTicks(handle.pid)` reads `/proc/777/stat` (`process-beast-executor.ts` ~559–563). Whether that read succeeds depends on whether the **runner host happens to have a live process with PID 777**:

- no real PID 777 → no ticks → stop options are `{ processGroupOwned: false }` → exact match passes;
- real PID 777 exists → ticks recorded (e.g. `"1155"`) → `processSignalOptions()` (~366–377) appends `expectedStartTimeTicks` → the exact-match assertion fails on the extra key.

That also explains the pattern: deterministic within one runner (retries see the same host state), intermittent across runs (different runners). The sibling identity-verification tests added in `d70b92ea` (#3441) already expect the `expectedStartTimeTicks` key (test file ~190–249); these two stop tests appear to have been left asserting the pre-#3441 exact shape.

## Possible fixes

- Assert with `expect.objectContaining({ processGroupOwned: ... })`, or build the expected object from `attempt.executorMetadata?.processStartTimeTicks` the way the executor does; or
- make the ticks source injectable/stubbable in tests so metadata for fake PIDs is deterministic (arguably nicer — `start()` reading `/proc` for a mocked PID is the underlying nondeterminism).

Happy to PR either variant if useful — but it may collide with whatever you have in flight, so flagging it first.

djm204:
## Summary
- use the guaranteed-live test runner PID in the two ProcessBeastExecutor operator-stop tests
- derive the exact expected signal options from persisted attempt identity metadata
- cover the `/proc` identity-token path deterministically instead of depending on whether host PID 777 exists

## Verification
- `npm test --workspace @franken/orchestrator -- --run tests/unit/beasts/process-beast-executor.test.ts` (55 passed)
- `npm run lint --workspace @franken/orchestrator -- --quiet`
- `npm run build`
- `npm run build --workspace @franken/orchestrator`

Closes #3576

### DIFF
--- a/packages/franken-orchestrator/tests/unit/beasts/process-beast-executor.test.ts
+++ b/packages/franken-orchestrator/tests/unit/beasts/process-beast-executor.test.ts
@@ -884,7 +884,7 @@ describe('ProcessBeastExecutor', () => {
     const repo = new SQLiteBeastRepository(join(workDir, 'beasts.db'));
     const logs = new BeastLogStore(join(workDir, 'logs'));
     const supervisor = {
-      spawn: vi.fn(async (_spec: unknown, _callbacks: unknown) => ({ pid: 777 })),
+      spawn: vi.fn(async (_spec: unknown, _callbacks: unknown) => ({ pid: process.pid })),
       stop: vi.fn(async () => {}),
       kill: vi.fn(async () => {}),
     };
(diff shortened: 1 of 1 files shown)


djm204:
@codex review

PR #3585, opened 2026-07-22
perf(planner): make topo sort queue linear

### LINKED ISSUE #3573: Planner topo sort uses O(n²) queue processing via Array.shift
## Description
**Quality goal:** Performance and resource bounds (`performance-resource-bounds`)

In `PlanGraph._kahn()` (topological sort), the queue of ready tasks is advanced with `queue.shift()` inside the main loop. This makes each dequeue O(n), so the total complexity of sorting a DAG can degrade to O(V²) for large graphs even when edges are sparse. That creates an unbounded CPU growth pattern relative to graph size, which conflicts with the performance-resource-bounds goal.

## Affected locations
- `packages/franken-planner/src/core/dag.ts:175-183` (`PlanGraph._kahn()`)
- `packages/franken-planner/src/core/dag.ts:169-172` (queue initialization and drain loop)

## Impact
For plans with many tasks, planner cycle checks and topological ordering become significantly slower as graph size grows, potentially stalling orchestration flows that repeatedly invoke `topoSort()`/`hasCycle()`. This can increase planning latency and event-loop contention in large-scale runs.

## Recommendation
Replace the dequeue pattern with an index-based queue (`let head = 0; while (head < queue.length) { const current = queue[head++]; ... }`) or a dedicated deque implementation so each pop is O(1), preserving overall O(V + E) behavior for Kahn traversal. Add a regression test that asserts complexity behavior with a larger synthetic dependency graph to prevent reintroduction.

djm204:
## Summary
- replace `Array.shift()` in Kahn traversal with an index-based queue head
- add a 1,024-task regression that detects linear-time front removal
- preserve topological ordering and cycle behavior while restoring O(V + E) traversal

## Verification
- `npm test` (planner): 248 passed
- `npm run typecheck` (planner)
- `npm run lint` (planner)
- `npm run build` (planner)

Closes #3573

### DIFF
--- a/packages/franken-planner/src/core/dag.ts
+++ b/packages/franken-planner/src/core/dag.ts
@@ -371,8 +371,10 @@ export class PlanGraph {
     }
 
     const sorted: TaskId[] = [];
-    while (queue.length > 0) {
-      const current = queue.shift() as TaskId;
+    let queueHead = 0;
+    while (queueHead < queue.length) {
+      const current = queue[queueHead] as TaskId;
+      queueHead += 1;
       sorted.push(current);
       for (const dependent of dependents.get(current) ?? []) {
         const newDeg = (inDegree.get(dependent) ?? 0) - 1;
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

PR #3911, opened 2026-07-30
fix(governor): authenticate approval responses before JSON parsing

### LINKED ISSUE #3722: Security: Authenticate approval responses before JSON parsing
## Description
**Quality goal:** Deterministic safety boundaries (`deterministic-safety`)

Owner: `packages/franken-governor` HTTP approval response maintainers.

The `/v1/approval/respond` public boundary parses the request body as JSON before the approval response has been authenticated with the governor signing secret. That means untrusted, unauthenticated input reaches `JSON.parse` before the deterministic safety boundary has verified the raw bytes with `verifyGovernorSignature`. The handler catches malformed JSON, but the enforcement ordering still violates the fail-closed rule that untrusted input is scanned/authenticated before parsing or dispatch.

## Affected locations
- `packages/franken-governor/src/server/app.ts:165-167` parses `rawBody` with `JSON.parse(rawBody.toString('utf8'))` before the later signature checks run.

## Impact
Unauthenticated callers can force the governor approval endpoint to parse arbitrary request bodies before signature verification. This weakens the approval boundary and leaves parsing behavior as a fallback path ahead of authentication, instead of failing closed on missing or invalid signatures first.

## Recommendation
Reorder `/v1/approval/respond` so it reads the raw bytes, verifies the configured governor signature against those bytes, and only then parses JSON. Preserve the explicit test-only unsigned path, but keep it gated before parsing as well. Add regression coverage proving unsigned, missing-signature, malformed-signature, and invalid-signature approval responses are rejected before JSON parsing or registry dispatch can occur.

djm204:
## Summary

- `POST /v1/approval/respond` in `packages/franken-governor/src/server/app.ts` parsed the raw request body with `JSON.parse` before the governor HMAC signature was verified against those bytes, so unauthenticated callers could force JSON parsing of arbitrary input ahead of the fail-closed auth boundary.
- Reordered the handler to authenticate the raw bytes first (via `verifyGovernorSignature`, or the explicit `allowUnsignedApprovalsForTests` test-only path, still gated ahead of parsing) and only parse JSON once authentication succeeds — matching the ordering already used by `POST /v1/approval/request`.
- No behavior change for authenticated callers; only the ordering of auth vs. parse changed.

## Test plan

- [x] Added TDD regression tests in `packages/franken-governor/tests/unit/server/app.test.ts` (`authenticates before JSON parsing` describe block) sending a malformed-JSON body under four cases — unsigned, missing signature, malformed signature, invalid signature — asserting each is rejected with the signature-specific 401 rather than the generic "Malformed JSON body" 400 that parsing failure would produce. Confirmed these 4 tests failed against the pre-fix ordering (red), then passed after the fix (green).
- [x] Added a 5th regression test proving registry dispatch never occurs when authentication fails, even with well-formed JSON.
- [x] `npx turbo run test --filter=@franken/governor` — 25 test files, 347 tests passing.
- [x] `npx turbo run typecheck --filter=@franken/governor` — clean.

Closes #3722

### DIFF
--- a/packages/franken-governor/src/server/app.ts
+++ b/packages/franken-governor/src/server/app.ts
@@ -430,13 +430,13 @@ export function createGovernorApp(options: GovernorAppOptions = {}): Hono {
   // POST /v1/approval/respond — respond to an approval request
   app.post('/v1/approval/respond', async (c) => {
     const rawBody = Buffer.from(await c.req.arrayBuffer());
-    const body = parseJsonBody(rawBody);
-    if (!body) {
-      return c.json({ error: { message: 'Malformed JSON body' } }, 400);
-    }
 
-    // Fail closed: unsigned approval responses are rejected unless a signing
-    // secret is configured (then verified) or an explicit test flag is set.
+    // Fail closed: authenticate the raw bytes with the governor signature
+    // BEFORE parsing JSON, so untrusted/unauthenticated input never reaches
+    // JSON.parse or registry dispatch. Unsigned approval responses are
+    // rejected unless a signing secret is configured (then verified) or an
+    // explicit test flag is set; the test-only unsigned path is still
+    // gated here, ahead of parsing.
     if (!options.signingSecret) {
       if (!options.allowUnsignedApprovalsForTests) {
         return c.json({ error: { message: 'Signing secret required for approval responses' } }, 401);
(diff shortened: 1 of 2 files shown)


djm204:
@codex review

